Issue #9, opened 2021-12-07
General Cleanup


StarArawn:
1. Cleanup code that is no longer used.
2. Cleanup examples
3. Cleanup warning messages.
4. Figure out if we are causing the vulkan validation errors/warnings or if its a bevy or wgpu issue. Example error:
```rust
ERROR wgpu_hal::vulkan::instance: VALIDATION [VUID-VkImageViewCreateInfo-image-04441 (0xb75da543)]
        Validation Error: [ VUID-VkImageViewCreateInfo-image-04441 ] Object 0: handle = 0x96fbe2000000005e, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0xb75da543 | Invalid usage flag for VkImage 0x96fbe2000000005e[] used by vkCreateImageView(). In this case, VkImage should have VK_IMAGE_USAGE_[SAMPLED|STORAGE|COLOR_ATTACHMENT|DEPTH_STENCIL_ATTACHMENT|INPUT_ATTACHMENT|TRANSIENT_ATTACHMENT|SHADING_RATE_IMAGE|FRAGMENT_DENSITY_MAP]_BIT set during creation. The Vulkan spec states: image must have been created with a usage value containing at least one of the usages defined in the valid image usage list for image views (https://vulkan.lunarg.com/doc/view/1.2.189.2/windows/1.2-extensions/vkspec.html#VUID-VkImageViewCreateInfo-image-04441)
```

StarArawn:
Fixed with the rewrite! 👍

PR #16, opened 2021-12-16
Cleanup 'bevy' example

ickk:
Relevant to issue #9.

Remove code duplication and demonstrate usage of the custom widget.


ickk:
The naming of `TestState` could be better. Perhaps 'ExampleWidget' or 'CustomWidget'?

PR #24, opened 2021-12-17
Add support for `MouseDown`, `MouseUp`, and `MousePressed` events

MrGVSV:
## Goal

Add `MouseDown`, `MouseUp`, and `MousePressed` events to listen for "half-clicks".

## Notes

This was done by changing the `previous_events` field to contain a `HashSet<EventType>` instead of a single `Option<EventType>`. Not sure how much that impacts performance.

Also, it's not the cleanest solution. We have to manually clear each event from the set whenever they no longer apply. For example, to clear `MouseUp` we wait for a `MouseDown`. I think to truly make this work nicely, we'd need an actual state machine.

But this could be a nice temporary solution until that's figured out.

### `MousePressed`

Currently set to run every frame. If this is too much of a performance hit, we can just get rid of it altogether, and leave managing that state up to the user.


StarArawn (commented):
On mobile at the moment but this looks good. I want to run it and give it a test before I approve/merge.

MrGVSV:
> On mobile at the moment but this looks good. I want to run it and give it a test before I approve/merge.

Definitely! I also started adding `MousePressed` but not sure if I'm happy with my implementation for that yet...

StarArawn (approved):

PR #25, opened 2021-12-17
Added text box..

StarArawn:
WIP


kayak_core/src/binding.rs
```diff
@@ -1 +1,29 @@
+use std::time::Instant;
+
 pub use flo_binding::{bind, notify, Binding, Bound, Changeable, MutableBound, Releasable};
+
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Debouncer {
+    last_updated: Instant,
+    time: f32,
```
MrGVSV, line 8: Personally, I think `threshold` or something similar might better communicate its purpose at a glance.

kayak_core/src/context.rs
```diff
@@ -1,18 +1,25 @@
...
+    multi_state::MultiState, widget_manager::WidgetManager, Event, EventType, Index, InputEvent,
+};
 
 pub struct KayakContext {
     widget_states: HashMap<crate::Index, resources::Resources>,
     global_bindings: HashMap<crate::Index, Vec<flo_binding::Uuid>>,
     widget_state_lifetimes:
         HashMap<crate::Index, HashMap<flo_binding::Uuid, Box<dyn crate::Releasable>>>,
-    current_id: Index,
+    pub current_id: Index,
```
MrGVSV, line 14: Should this actually be public? The `set_current_id` has a bit of extra logic that could be bypassed with this.

https://github.com/StarArawn/kayak_ui/blob/772a6d79c7d7176acb3012258fd39441b183ab0b/kayak_core/src/context.rs#L102-L106

kayak_widgets/src/text_box.rs
```diff
@@ -0,0 +1,105 @@
...
+    let background_styles = Style {
+        background_color: StyleProp::Value(Color::new(0.176, 0.196, 0.215, 1.0)),
+        border_radius: StyleProp::Value((5.0, 5.0, 5.0, 5.0)),
+        height: StyleProp::Value(Units::Pixels(26.0)),
+        padding_left: StyleProp::Value(Units::Pixels(5.0)),
+        padding_right: StyleProp::Value(Units::Pixels(5.0)),
+        ..styles.clone().unwrap_or_default()
+    };
+
+    let internal_value = context.create_state("".to_string()).unwrap();
```
MrGVSV, line 56: Shouldn't this be:

```suggestion
    let internal_value = context.create_state(value.clone()).unwrap();
```

to account for the `value` starting as a non-empty String?

MrGVSV, line 56: Also, is the internal state necessary? Is there a reason we don't just use `value` directly?

With `value` and `on_change`, I imagine this is a controlled state, meaning nothing should update unless the caller says so, right? They could theoretically allow their `on_change` callback to not do anything, or even replace `value` with a new String.

kayak_core/src/context.rs
StarArawn: I exposed for testing, but I'll revert that change.

kayak_core/src/binding.rs
StarArawn: Yeah I ended up not needing it, but I will change it to threshold because that makes sense. 👍

kayak_widgets/src/text_box.rs
StarArawn: yeah that's a good point 👍

StarArawn: The API's changed a bit here and create_state is just legacy stuff. I think I can remove it.

PR #28, opened 2021-12-18
Fixed backspace for textbox input

MrGVSV:
Textbox wasn't checking for `'0x7f'` (another common backspace sequence). Added a check for that and extracted it to a separate function for clarity.


MrGVSV:
Although, would it be better to wrap the `char` in the `CharInput` event with a struct that contains methods for checking things like `is_shift(&self)`, `is_return(&self)`, etc.? And then `impl From<CharWrapper> for char` so we can just use it as a `char` if we want to.

Then again, I guess that could just be done with `KeyCode`, so maybe not needed 🤔

PR #29, opened 2021-12-19
Fully working todo..

StarArawn:
Many bug fixes...

PR #31, opened 2021-12-19
Extracted ChangeEvent and made it generic

MrGVSV:
## Goal

The `ChangeEvent` from `kayak_widgets/src/text_box.rs` could be useful for callbacks in other widgets. This logic could be moved to `kayak_core` so that it may be used in other scenarios.

## Solution

Extracted the `ChangeEvent` from `kayak_widgets/src/text_box.rs` and added it to `kayak_core`.

Also made the event generic so that values of any type could be sent (e.g. `ChangeEvent<bool>`, `ChangeEvent<MyStruct>`, etc.).

## Notes

This might change if props become restricted to `Binding<T>` types, but could be a decent addition in the meantime.

### DIFF
--- a/examples/text_box.rs
+++ b/examples/text_box.rs
@@ -3,13 +3,14 @@ use bevy::{
     window::WindowDescriptor,
     DefaultPlugins,
 };
+use kayak_core::OnChange;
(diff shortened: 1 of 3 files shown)


MrGVSV:
Closing due to this being possibly being added in #29.

PR #32, opened 2021-12-20
Add: `Fold` widget

MrGVSV:
## Goal

Encapsulate the common use-case of a foldout/dropdown container.

## Solution

Added the `Fold` widget to allow such a container to be defined like:

```rust
rsx! {
  <Fold label={"Some Label".to_string()} open={false} on_change={None}>
    // ...content...
  </Fold>
}
```

## Example
fold-example

## Areas of Improvement

This widget could be improved a few ways:

* Adding a prop to set the left-side margin automatically on children
* Separating un-controlled (`default_open`) and controlled (`open`/`on_change`)
* Make the label focusable
* Add animations for opening and closing

> The first two likely rely on a change to make widgets have default props.

## Tasks
- [x] Add `Fold` widget
- [x] Add example showcasing the widget
- [x] Wait for generic `OnChange` to arrive from #29 (this PR currently uses its own `OnChange`)


kayak_widgets/src/fold.rs
```diff
@@ -0,0 +1,107 @@
...
+///     <Fold label={"Toggle Open/Close".to_string()} open={open} on_change={Some(on_change)}>
+///         <Text content={"Fold Content".to_string()} size={16.0} />
+///     </Fold>
+/// }
+/// ```
+#[widget]
+pub fn Fold(label: String, children: Children, open: bool, on_change: Option<OnChange<bool>>) {
+
+    // === State === //
+    let open_clone = open.clone();
```
StarArawn, line 38: We can get rid of some of these clones now.

StarArawn (approved):

PR #39, opened 2021-12-22
Added Provider/Consumer system

MrGVSV:
## Objective

Create a way for widgets to pass data to descendants without having to pass them as props to each node down to the consuming widget.

## Solution

Taking the [Context](https://reactjs.org/docs/context.html) approach from React, added `create_provider` and `create_consumer` methods to `KayakContext`. Using this allows for widgets to either provide state data to their children or for widgets to consume data from a parent provider.

### Example

```rust
#[widget]
fn MyProvider() {
  let data: Binding<Foo> = context.create_provider(Foo { bar: 123 });
  let foo = data.get();
  // ...
}

#[widget]
fn MyConsumer() {
  let data: Binding<Foo> = context.create_consumer::<Foo>().unwrap();
  let foo = data.get();
  // ...
}

#[widget]
fn MyWidget() {
  // ...
  rsx! {
    <MyProvider>
      // ...
      <MyConsumer />
      // ...
    </MyProvider>
  }
}
```

## Tasks

- [x] Add `create_provider` and `create_consumer` methods
- [x] Create `TooltipProvider` and `TooltipConsumer` widgets to showcase this new system
  - [x] Include them in an example (or create a new one)
- [x] Create an example showing how to use the provider/consumer system

## Demo

From the added `provider.rs` example. This shows how to implement the provider/consumer pattern for theming (as opposed to using a global state) and how to nest providers to allow for multiple themes.

provider-demo


kayak_core/src/context.rs
```diff
@@ -102,6 +99,66 @@ impl KayakContext {
...
+        provider.insert(state.clone());
+        providers.insert(self.current_id, provider);
+
+        state
+    }
+
+    /// Creates a consumer for the given type, [T]
+    ///
+    /// This allows direct access to a parent's state data made with [create_provider](Self::create_provider).
+    pub fn create_consumer<T: resources::Resource + Clone + PartialEq>(&mut self) -> Option<Binding<T>> {
```
MrGVSV, line 136: Should this be called something else? I debated calling it `get_provider`. 

They could even be named something else if there's something better. I chose provider/consumer since it's what React has, but I'm open to suggestions!

kayak_core/src/context.rs
```diff
@@ -102,6 +99,66 @@ impl KayakContext {
...
+
+        let mut provider = resources::Resources::default();
+        let state = crate::bind(initial_state);
+        let lifetime = Self::create_lifetime(&state, &self.widget_manager, self.current_id);
+        Self::insert_state_lifetime(
+            &mut self.widget_state_lifetimes,
+            self.current_id,
+            state.id,
+            lifetime,
+        );
```
MrGVSV, line 126: The providers are the only thing being bound to the state. If we make it so that children don't necessarily always update when the parent does, then we can add this logic to the `create_consumer` method as well.

kayak_core/src/context.rs
```diff
@@ -536,6 +585,13 @@ impl KayakContext {
         }
     }
 
+    /// Get the last calculated mouse position.
+    ///
+    /// Calling this from a widget will return the last mouse position at the time the widget was rendered.
+    pub fn last_mouse_position(&self) -> (f32, f32) {
+        self.last_mouse_position
+    }
```
MrGVSV, line 593: Should this be exposed? I added it so that the tooltips could access the cursor position when they re-render.

I would put this data on hover, but it doesn't work well with the current event system sine `EventType::Hover {pos: (123.0, 123.0)}` is different from `EventType::Hover {pos: (456.0, 456.0)}` And we need those to be hashable so we can look them up relatively easily.

src/widgets/tooltip.rs
```diff
@@ -0,0 +1,213 @@
...
+///                 // ...
+///             </TooltipConsumer>
+///             // ...
+///         </TooltipProvider>
+///     </>
+///   }
+/// }
+/// ```
+#[widget]
+pub fn TooltipProvider(
```
MrGVSV, line 68: Should tooltips be removed as a widget or are they okay to keep?

There's a few things they need to be _mostly_ feature-complete:
1. Dynamic resizing based on text (is there a way to get this info on render?)
2. Nesting of consumers (hover events are called on every node that the cursor is hovered on, whether on top or not, preventing this from working)
3. Some layout issues regarding `right` and `bottom` (might be an OS issue?)

They're good enough to work with, but no where near what they should be as of yet.

kayak_core/src/context.rs
```diff
@@ -102,6 +99,66 @@ impl KayakContext {
...
+            state.id,
+            lifetime,
+        );
+        provider.insert(state.clone());
+        providers.insert(self.current_id, provider);
+
+        state
+    }
+
+    /// Creates a consumer for the given type, [T]
```
StarArawn, line 133: I like the Consumer/Provider names. Can we perhaps mention that its a context? Something like:
```suggestion
    /// Creates a context consumer for the given type, [T]
```

kayak_core/src/context.rs
```diff
@@ -102,6 +99,66 @@ impl KayakContext {
...
+        let type_id = std::any::TypeId::of::<T>();
+
+        let providers = if let Some(providers) = self.widget_providers.get(&type_id) {
+            providers
+        } else {
+            // Early escape
+            return None;
+        };
+
+        let mut index = Some(self.current_id);
```
StarArawn, line 146: I think you can move this code into the if statement and avoid the early escape.
```suggestion
        if let Some(providers) = self.widget_providers.get(&type_id) {
            let mut index = Some(self.current_id);
            while index.is_some() {
                // Traverse the parents to find the one with the given state data
                index = self.widget_manager.tree.get_parent(index.unwrap());
    
                let key = index.unwrap();
                if let Some(provider) = providers.get(&key) {
                    if let Ok(state) = provider.get::<Binding<T>>() {
                        return Some(state.clone());
                    }
                }
            }
        }
```

kayak_core/src/context.rs
StarArawn: I think doing it this way is fine for now. 👍

src/widgets/tooltip.rs
```diff
@@ -0,0 +1,213 @@
...
+use crate::widgets::{Background, Clip, Element, If, Text};
+
+#[derive(Clone, PartialEq, Debug, Default)]
+pub struct TooltipData {
+    /// The anchor coordinates in pixels (x, y)
+    ///
+    /// If `None`, defaults to cursor position
+    pub anchor: Option<(f32, f32)>,
+    /// Used to force an update
+    ticker: u8,
```
StarArawn, line 17: Does it not update when any of the other fields here change?

src/widgets/tooltip.rs
StarArawn: > Should tooltips be removed as a widget or are they okay to keep?

They are fine to keep!

> Dynamic resizing based on text (is there a way to get this info on render?)

Not yet but I will be adding a measure_text possibly to the context?

> Nesting of consumers (hover events are called on every node that the cursor is hovered on, whether on top or not, preventing this from working)

We can perhaps add a `hoverable` similarly to how we handled `focusable`?

> Some layout issues regarding right and bottom (might be an OS issue?)

Layout issues with text?

src/widgets/tooltip.rs
```diff
@@ -0,0 +1,213 @@
...
+        }
+        _ => {}
+    }));
+
+    rsx! {
+        <>
+            {children}
+        </>
+    }
+}
```
StarArawn, line 213: New line please!

StarArawn (changes requested):
Overall this PR is looking good. Just a few small changes requested. I really like the theme demo its cool to see it in action!

src/widgets/tooltip.rs
MrGVSV: > Not yet but I will be adding a measure_text possibly to the context?

That would be very helpful. Is there any plans to allow widgets to access their cached layout? Sorta like [`onLayout`](https://reactnative.dev/docs/view#onlayout) from React Native.

> We can perhaps add a `hoverable` similarly to how we handled `focusable`?

Maybe. This would help solve the issue of what's being hovered. Although, could we also achieve this by adding a way to set `should_propagate` to `false`? Then we can propagate it like `Text` → `Clip` → `Background` → `MyWidget` and handle the event on `MyWidget` (or just prevent it from bubbling up to the parent).

But I think we also need a way to only mark only the topmost node as "hovered". If a window is on top of another window, both containing buttons, then both buttons might be marked as "hovered" when really we only want the button from the topmost window. Is there a way we could do this?

I also may have no clue what I'm talking about haha

> Layout issues with text?

No, with the containing `Background` (and all other widgets). Maybe I'm doing something wrong, but `right` and `bottom` position the tooltip stuck to the edge of the screen. And (maybe related) `padding_*`, and `margin_*` seem to not work at all for me.

I thought it might be an OS issue or another DPI issue because my widgets are all cutoff at 50% width and height of the physical window. So maybe that's also affecting layout? I'll open an issue about it later tonight with more details.

examples/provider.rs
```diff
@@ -0,0 +1,274 @@
...
+            <ThemeSelector />
+
+            <Background styles={Some(bg_style)}>
+
+                <Text content={text} size={12.0} styles={Some(text_style)} />
+                <Background styles={Some(btn_style)}>
+                    <Text content={btn_text} size={14.0} styles={Some(btn_text_style)} />
+                </Background>
+
+                <If condition={is_root}>
```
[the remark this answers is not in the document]
MrGVSV: Yeah! Although a better way would probably to use a `u8` and just pass it to the next recursion but with `1` subtracted. But I thought that was a bit overkill for this example haha

kayak_core/src/context.rs
```diff
@@ -1,13 +1,15 @@
-use flo_binding::Changeable;
+use flo_binding::{Binding, Changeable};
```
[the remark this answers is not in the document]
MrGVSV: Makes sense! I'll check the other files too and make sure I don't miss any.

kayak_core/src/context.rs
```diff
@@ -102,6 +99,66 @@ impl KayakContext {
...
+            if let Ok(state) = provider.get::<Binding<T>>() {
+                // Provider was already created
+                return state.clone();
+            }
+        }
+
+        let mut provider = resources::Resources::default();
+        let state = crate::bind(initial_state);
+        let lifetime = Self::create_lifetime(&state, &self.widget_manager, self.current_id);
+        Self::insert_state_lifetime(
```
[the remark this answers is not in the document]
MrGVSV: Yeah I was curious about when these get removed too

kayak_core/src/context.rs
MrGVSV: I wanted to do this, but I was hesitant to use the term "context" because I was worried about causing confusion with `KayakContext`. But I can definitely add this terminology across the documentation.

src/widgets/tooltip.rs
MrGVSV: It does. Sorry, this might actually not be needed anymore. I can remove it and the `mark_dirty` method.

I originally had to do this because I set the `anchor` directly on `TooltipProvider` instead of going through the `TooltipConsumer`. And we needed a way of telling `TooltipProvider` to update when the cursor moved (since the other fields didn't change).

src/widgets/tooltip.rs
StarArawn: > No, with the containing Background (and all other widgets). Maybe I'm doing something wrong, but right and bottom position the tooltip stuck to the edge of the screen. And (maybe related) padding_*, and margin_* seem to not work at all for me.

Ah yeah I think I have to remove margin I was miss-understanding as to how morphorm works with margins. You have to use left, right, top, bottom to add margin.

MrGVSV: Oh okay gotcha. That's what I ended up doing anyways so that works out! 😁

StarArawn (approved):

PR #40, opened 2021-12-22
Measure text

StarArawn:


src/widgets/text.rs
```diff
@@ -1,18 +1,50 @@
...
-    styles::{Style, StyleProp},
+    styles::{Style, StyleProp, Units},
     widget,
 };
 
 #[widget]
 pub fn Text(size: f32, content: String, styles: Option<Style>, font: Option<u16>) {
+    let px_size = {
+        let mut px_size = (0.0, 0.0);
+        // TODO: This is super gross! I think we should pass in fonts to the context and let it keep track of fonts on its own for text measuring.
```
MrGVSV, line 11: I agree that this should probably be managed by the context.

It's not ideal, but we could maybe block while we load the font? Or possibly require the user to set the font with the already loaded asset, kinda like [`TextureAtlasBuilder`](https://github.com/bevyengine/bevy/blob/22c665fa39b16c62272b927db742ea62d67615e8/examples/2d/texture_atlas.rs#L43-L49).

In either case, it would probably be better to keep main functionality like font management within Kayak so it could potentially be used apart from Bevy (if that's even your goal).

PR #43, opened 2021-12-27
Re-implemented event system

MrGVSV:
## Objective

The current event system is limited in that it:
* Does not properly target nodes
* Does not support event propagation

For example, in the following tree:

```
+-- Window
|  +-- Element
|  |  +-- Button
```

Hovering over the `Button` will also count as hovering over the `Element` and `Window`.

We also can't just filter by z-index or tree depth alone, since events like `MouseIn` should apply over all appropriate nodes in the tree (whether stacked atop each other or not).

## Solution

With this PR, we now find the target node rather than sending the event to every node along the way to the actual target. Doing this allows us to better control which nodes receive the event. It also allows us to properly propagate the event up the tree and stop propagation when requested.

### Issues

In order to properly find the target of a mouse event, we need to look through every node. This is due to issues regarding self-directed and overflowing elements. This prevents us from recursively narrowing down on the target widget, since we don't know if the parent actually contains it or not.

I haven't tested it on a huge tree, so it might not be a _huge_ issue, but it's not great. If this is too big of an issue, then maybe a different method could be considered. One fix could be to revert back to the previous system and combine it with the functionality of `WidgetManager::get_nodes_under`. Doing that should allow us to do everything in a single iteration of the node tree (hopefully?).

Marking this as a draft for now until I can take a closer look at this potential fix.

---

**Edit:** After trying out the idea of performing all checks within a single tree iteration, I managed to get it working but found one small issue. Bevy might send multiple `CursorMoved` events in a single frame. If we were to process each one, we'd have to add some complicated logic to switch between active `current_mouse_position` values (which, in any case, may be incorrect for other non-`CursorMoved` events on the same widget anyways due to event ordering).

Therefore, I opted to skip all but the last `CursorMoved` event. This means we'll miss some for really fast cursor movements, but I think it should generally be okay. Let me know if this is not an ideal solution, though!


kayak_core/src/widget_manager.rs
```diff
@@ -384,4 +385,97 @@ impl WidgetManager {
...
+    /// Get the nodes under the given position
+    ///
+    /// The first element in the returned tuple is the node deemed as the "best match". A node is a best match when
+    /// it is the deepest matching node in the hierarchy or it has a higher z-index than the other matches.
+    ///
+    /// # Arguments
+    ///
+    /// * `position`: The 2D point to check under
+    /// * `parent`: The parent to start at (or `None` for the root element)
+    /// * `respect_pointer_events`: Whether a widget's `pointer_events` style should be respected
```
MrGVSV, line 398: Not a huge fan of adding this parameter. Another reason to maybe move this logic to `KayakContext::process_events`.

kayak_core/src/widget_manager.rs
```diff
@@ -384,4 +385,97 @@ impl WidgetManager {
...
+    ///
+    pub fn get_nodes_under(&self, position: (f32, f32), parent: Option<Index>, respect_pointer_events: bool) -> Option<(Index, Vec<Index>)> {
```
MrGVSV, line 400: Also not a fan of this return data

kayak_core/src/event.rs
```diff
@@ -25,9 +56,57 @@ pub enum EventType {
...
+            Self::Click => true,
+            Self::MouseDown => true,
+            Self::MouseUp => true,
+            Self::CharInput { .. } => true,
+            Self::KeyboardInput { .. } => true,
+            // Doesn't Propagate
+            Self::MouseIn => false,
+            Self::MouseOut => false,
+            Self::Focus => false,
+            Self::Blur => false,
```
MrGVSV, line 90: Verbose so that nothing gets missed if we add/remove events.

kayak_core/src/context.rs
```diff
@@ -267,222 +267,200 @@ impl KayakContext {
...
+    ///
+    /// ## Target:
+    ///   The Target Phase simply identifies the target for an event so that we can generate the propagation path
+    ///   for it.
+    ///
+    /// ## Propagate:
+    ///   The Propagate Phase (also known as the Bubble Phase) is where we bubble up the tree from the target node,
+    ///   firing the bubbled event along the way. At any point, the bubbling can be stopped by calling
+    ///   [`event.stop_propagation()`](Event::stop_propagation). Not every event can be propagated, in which case,
+    ///   they will only fire for their specified target.
```
MrGVSV, line 289: Can remove this from the documentation and just put it as a separate comment if that is preferred.

MrGVSV:
The latest change now moves all of the event management to a separate `EventDispatcher` struct on `KayakContext`. This was mainly a code organization thing. But doing this also makes it much easier to control what event data is shared with the context (and hopefully avoid confusion on the purpose of certain fields when contributing to `KayakContext`).

If this is too big a change or something that isn't wanted, let me know, and I can revert it!

StarArawn (approved):
This cleans things up quite a bit! I think I like everything except the part about searching every node. :( Can we add an issue and perhaps in the future we can fix that? I'm fine with merging this as is for now though.

MrGVSV:
> This cleans things up quite a bit! I think I like everything except the part about searching every node. :( Can we add an issue and perhaps in the future we can fix that? I'm fine with merging this as is for now though.

Thanks! Yeah I’m not a fan either. Thankfully I did manage to get it down from searching every node on _every event_, to just a single search, like we had before with `down_iter`. But I think we can do better if we can figure out a clever way of determining which children might contain a point for overflowing/self-directed nodes. But yeah, that can be figured out later haha, so I'll just open an issue about it for now.

PR #45, opened 2021-12-28
Added `use_effect` hook macro

MrGVSV:
## Objective

Add an easy way for callbacks to be invoked when a state changes as a side-effect or when a widget first renders (then never again).

We can't simply call `my_state.when_changed(notify(/* ... */)).keep_alive()` because this will create an additional callback on every re-render. And we can't pass the `Releasable` into `use_state!` either because it doesn't implement `Clone`.

## Solution

Taking again from [React](https://reactjs.org/docs/hooks-effect.html), this PR adds a system for creating side-effects and an accompanying `use_effect` macro hook. 

This new system stores closures in `KayakContext` and registers it to be called whenever an object implementing `Changeable` (aka `Binding`) in an array of them (the dependencies) changes.

### Examples

You can respond to one state change and change another state in response:

```rust
use_effect!(move || {
  set_other_state(count_state.get() * 2);
}, [count_state]);
```

They can also just be used to run a bit of logic a single time (upon first render):

```rust
use_effect!(|| {
  // Do something...
}, []);
```

### Cleanup

Unlike React's `useEffect` hook, we don't currently have a way to handle logic for unmounting. This could later be added though by creating the side-effect with a `computed` binding.

---

I also added a new example dedicated to the `use_state`, `use_effect`, and other future hooks.

### DIFF
--- a/kayak_render_macros/src/lib.rs
+++ b/kayak_render_macros/src/lib.rs
@@ -114,24 +114,27 @@ pub fn dyn_partial_eq(_: TokenStream, input: TokenStream) -> TokenStream {
     .into()
 }
 
-/// Create a state and its setter
+/// Register some state data with an initial value.
+///
+/// Once the state is created, this macro returns the current value, a closure for updating the current value, and
+/// the raw Binding in a tuple.
 ///
 /// For more details, check out [React's documentation](https://reactjs.org/docs/hooks-state.html),
 /// upon which this macro is based.
 ///
 /// # Arguments
 ///
-/// * `initial_state`: The expression
+/// * `initial_state`: The initial value for the state
 ///
-/// returns: (state, set_state)
+/// returns: (state, set_state, state_binding)
 ///
 /// # Examples
 ///
 /// ```
 /// # use kayak_core::{EventType, OnEvent};
-/// use kayak_render_macros::use_state;
+/// # use kayak_render_macros::use_state;
 ///
-/// let (count, set_count) = use_state!(0);
+/// let (count, set_count, ..) = use_state!(0);
 ///
 /// let on_event = OnEvent::new(move |_, event| match event.event_type {
 ///         EventType::Click => {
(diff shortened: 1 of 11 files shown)


kayak_render_macros/src/lib.rs
```diff
@@ -145,8 +150,23 @@ pub fn dyn_partial_eq(_: TokenStream, input: TokenStream) -> TokenStream {
...
+    }};
+    TokenStream::from(result)
+}
+
+/// This macro works exactly like its [use_state] counterpart, except that it also returns
+/// the raw state Binding as the third field in the tuple.
+///
+/// returns: (state, set_state, state_binding)
+#[proc_macro]
+pub fn use_raw_state(initial_state: TokenStream) -> TokenStream {
```
MrGVSV, line 166: I'm debating on this. I added `use_raw_state` so we could access the actual state `Binding`, for use in the `use_effect` hook. But we could also just add that to the existing `use_state` hook, which could be cleaner.

However, doing this would require we create states like this, when we don't care about the raw `Binding`:

```rust
// We now have to add the ", .." part
let (state, set_state, ..) = use_state!(0usize);
```

I don't think it's the worst, but I'm not sure. Any thoughts?

MrGVSV:
Also, if you'd prefer a different name (`use_closure`, `use_detector`, etc.) in place of `use_effect`, just let me know!

I chose it because of React's convention, and I tried my best to explain what a side-effect is in simple terms. Hopefully those things help newer devs avoid any sort of confusion 😅

kayak_render_macros/src/use_effect.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /// Build the output token stream, creating the actual use_effect code
+    pub fn build(self) -> TokenStream {
+        let dep_array = self.create_dep_array();
+        let cloned_deps = self.create_clone_deps();
+        let closure = self.closure;
+        let result = quote! {{
+            use kayak_core::{Bound, MutableBound};
+            #cloned_deps;
```
MrGVSV, line 59: Maybe I should also clone the dependencies (without the custom naming) so users don't have to do it themselves when using a state internally in the closure. Then again that might be unintuitive and would probably contain unnecessary clones? But it could also streamline the usage? Hm 🤔

kayak_render_macros/src/lib.rs
StarArawn: I'm wondering if it might be better to destructure the binding somehow. 🤔 something like:
```rust
let binding = context.create_state(0usize);
let (state, set_state) = binding.get_state();
```

I think I prefer this though:
```rust
let (state, set_state, ..) = use_state!(0usize);
```

kayak_render_macros/src/use_effect.rs
StarArawn: I think this looks fine as is.

kayak_render_macros/src/lib.rs
MrGVSV: Ooh that would nice. We could maybe do that with a newtype around `Binding`, then impl `Deref` and stuff to access the raw `Binding`?

Then we wouldn't even really need the `use_state` macro which might be nice.

But yeah for now, I think I prefer 86ing the `use_raw_state`. I'll move the logic over to `use_state`.

StarArawn: yeah I want to move flo_binding directly into this project since it doesn't seem like its updated anymore.

MrGVSV: Yeah that would be really nice. There are some things that could be more specialized for Kayak's purposes.

StarArawn (approved):

PR #48, opened 2022-01-04
Dont require startup_system, make BevyContext optional

RobertoSnap:
Discord: https://discord.com/channels/691052431525675048/742884593551802431/927920182830194688

### DIFF
--- a/bevy_kayak_ui/src/lib.rs
+++ b/bevy_kayak_ui/src/lib.rs
@@ -36,18 +36,18 @@ pub(crate) fn to_bevy_color(color: &kayak_core::color::Color) -> Color {
 }
 
 pub fn update(world: &mut World) {
-    let bevy_context = world.remove_resource::<BevyContext>().unwrap();
-    if let Ok(mut context) = bevy_context.kayak_context.write() {
-        context.set_global_state(std::mem::take(world));
-        context.render();
-        *world = context.take_global_state::<World>().unwrap()
+    if let Some(bevy_context) = world.remove_resource::<BevyContext>() {
+        if let Ok(mut context) = bevy_context.kayak_context.write() {
+            context.set_global_state(std::mem::take(world));
+            context.render();
+            *world = context.take_global_state::<World>().unwrap()
+        }
+        world.insert_resource(bevy_context);
     }
-
-    world.insert_resource(bevy_context);
 }
 
 pub fn process_events(
-    bevy_context: Res<BevyContext>,
+    bevy_context: Option<Res<BevyContext>>,
     windows: Res<Windows>,
     mut cursor_moved_events: EventReader<CursorMoved>,
     mut mouse_button_input_events: EventReader<MouseButtonInput>,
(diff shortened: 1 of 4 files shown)


RobertoSnap:
Added a test in Kayak bevy_conditiona.rs where it changes from add_startup_system() to add_system().

Getting error:
`thread 'Compute Task Pool (1)' panicked at 'A node_tree should have a root node', kayak_core/src/widget_manager.rs:328:18`

https://github.com/StarArawn/kayak_ui/blob/a4bdcbe5e259456d4c5617bb2e0b7443872d472e/kayak_core/src/widget_manager.rs#L326

Cant see why this is happening yet

StarArawn:
> Added a test in Kayak bevy_conditiona.rs where it changes from add_startup_system() to add_system().
> 
> Getting error: `thread 'Compute Task Pool (1)' panicked at 'A node_tree should have a root node', kayak_core/src/widget_manager.rs:328:18`
> 
> https://github.com/StarArawn/kayak_ui/blob/a4bdcbe5e259456d4c5617bb2e0b7443872d472e/kayak_core/src/widget_manager.rs#L326
> 
> Cant see why this is happening yet

Check out this system here:
https://github.com/StarArawn/kayak_ui/blob/main/bevy_kayak_ui/src/render/unified/mod.rs#L69

MrGVSV:
> Added a test in Kayak bevy_conditiona.rs where it changes from add_startup_system() to add_system().
> 
> Getting error: `thread 'Compute Task Pool (1)' panicked at 'A node_tree should have a root node', kayak_core/src/widget_manager.rs:328:18`
> 
> https://github.com/StarArawn/kayak_ui/blob/a4bdcbe5e259456d4c5617bb2e0b7443872d472e/kayak_core/src/widget_manager.rs#L326
> 
> Cant see why this is happening yet

My guess would be that this is a system ordering error (i.e. renderer tries to build primitives before the tree is built). But I could be wrong.

RobertoSnap:
I think my knowledge of Bevy and Kayak is lacking to be able to solve this one.

I see that the root node is None, but i dont understand the flow good enough yet to be able to pinpoint why and how to work with it.

CleanShot 2022-01-05 at 12 19 23@2x

RobertoSnap:
Its looks like widget is NOT added, could it be that the widget macro has not propegated or something?

MrGVSV:
> I think my knowledge of Bevy and Kayak is lacking to be able to solve this one.
> 
> I see that the root node is None, but i dont understand the flow good enough yet to be able to pinpoint why and how to work with it.
> 
> CleanShot 2022-01-05 at 12 19 23@2x
> 
> Learnt a great lot thought.

Your debug point is set before the call to initialize the the context. Could you put a `println!` or another debug point in `WidgetManager::create_widget` to see if the widgets are in fact never added?

RobertoSnap:
CleanShot 2022-01-05 at 19 38 22

But the Widget is created, thought without a parent.
CleanShot 2022-01-05 at 19 40 30

MrGVSV:
That's probably the `<App />` widget. And do any of the `<Window />` widgets get added?

RobertoSnap:
Nope, it fails at App widget -> build_render_primitives (see id 0 in image) because it does not have root node. ( it seems like)
CleanShot 2022-01-05 at 20 13 17

MrGVSV:
Not sure if this is a solution, but what happens if you move the check for root outside the call to `Self::recurse_node_tree_to_build_primitives`? And if the root node doesn't exist, return an empty `Vec`

RobertoSnap:
CleanShot 2022-01-05 at 20 50 53

It runs, but nothings gets rendered. 

Could it be some that some system is dependent on a root node added in add_startup_system()?
CleanShot 2022-01-05 at 20 57 17

I just tried something (but dont know what im doing TBH) :
CleanShot 2022-01-05 at 21 05 28


CleanShot 2022-01-05 at 21 03 57

So it has managed to create some context ( i dont know what a valid one looks like), but nothing gets rendered.

Sorry, im fumbling. I probably need to up my Bevy and Rust skills first :)

MrGVSV:
> Sorry, im fumbling. I probably need to up my Bevy and Rust skills first :)

No worries, this is probably one of the best paths to learning anyways :)

StarArawn:
> So it has managed to create some context ( i dont know what a valid one looks like), but nothing gets rendered.

What does custom widget look like? You can put a dbg! statement in the extract function of the renderer to see what its passing to the renderer.

Issue #41, opened 2021-12-23
Add `rustfmt.toml`


MrGVSV:
To better maintain style consistency within the codebase, a `rustfmt.toml` file should be added to allow automatic formatting of code.

StarArawn:
Code auto formats for me on save, but yes rustfmt.toml would help! 👍

IceSentry:
A rustfmt.toml isn't required if we use the default rustfmt configs, but it could be added as a CI step to enforce it on PRs

StarArawn:
CI is checking formatting now. Do we still need a rustfmt.toml? I think we can just keep using the default config.

IceSentry:
I don't think a rustfmt.toml is necessary if you are fine with the default config.

MrGVSV:
I think default should be fine for now.

PR #50, opened 2022-01-07
Improve Focus Management

MrGVSV:
## Objective

Add better focus management so we can navigate between focusable widgets (good for accessibility).

## Solution

The main solution was the addition of a `FocusTree` struct which cached the "focusable" widgets so navigation between them didn't result in possibly iterating over the whole tree (at worst). This also allows the navigation between these focusable widgets to cycle once an end is reached.

Additionally, this resulted in a number of other small additions/changes:

* The ability to run default kayak-defined default actions on an event. This allows us to do things like _"navigate to next focus on <kbd>Tab</kbd>."_
  * Also added the option to prevent these default actions by calling `Event::prevent_default()`.
  * Current default actions:
    * Navigate to next focusable widget on <kbd>Tab</kbd>
    * Navigate to previous focusable widget on <kbd>Tab</kbd> + <kbd>Shift</kbd>
* `Widget::focusable()` now returns `Option<bool>`. This is used by the `FocusTracker` struct to see if a widget is "focusable." `None` means it has no preference, while `Some(true)` or `Some(false)` allow or disallow focusability, respectively.
  * The focusability is tracked from props set by the parent and from the widget itself (i.e. in a callback). Focusability from the widget itself overrides whatever is set by the parent (including `#[widget(focusable)]`).
* New methods on `Tree` for simpler usage. Some of the usages for these were removed, but I left the code in, just in case we want them (and I can just as easily remove them if preferred).
  * `remove(...)` - Remove the given node and its descendants
  * `remove_and_reparent(...)` - Remove the given node and reparent its children
  * `replace(...)` - Replace the given node with another, transferring children
  * `contains(...)` - Checks if the given node is contained within the tree
  * `len()` - Returns the number of nodes in the tree
  * `is_empty()` - Checks if the tree is empty
  * `is_descendant(...)` - Checks if the given node is a descendant of another node
  * Added tests to help ensure all these methods work as intended
* Implemented `Tree::get_last_child(...)`
* Fixed `Tree::get_prev_sibling(...)`
* Made `Button` focusable by default

## Tasks

- [x] Add `FocusTree` struct to manage focus
- [x] Integrate focus with event system
  - [x] Allow default actions to be defined and prevented
  - [x] Add a default action that navigates to next focusable widget on <kbd>Tab</kbd>
  - [x] Add a default action that navigates to previous focusable widget on <kbd>Shift</kbd> + <kbd>Tab</kbd>
  - [x] ~~Fix: Tabbing should resume from last focused widget after blur~~ This behavior actually appears to be standard on the web, so maybe we'll keep it for now
- [x] Add ability to toggle focusability (if possible)
  - [x] Add/Remove widgets when changed (needs fixing - see notes)
- [x] Add `tab_box` example utilizing this new system

### Known Issues

Currently, this solution rebuilds the focus tree every render. This is not ideal, but it helps keep things clean. The reason we need to do this, is because we don't have a great way of running code when a widget is removed (not without iterating through all the changes every render). If this is sorted out in a clean way, we can move the logic over to there and add to or remove from the focus tree only when needed, rather than clearing and rebuilding every frame.

Another downside is that the rebuilding code is integrated into the node building code, which iterates through the widgets in order of creation (for performance reasons). This means that when clearing the focus tree and adding the widgets back in, we are not respecting their order in the tree (i.e. siblings can be inserted out of order).

The added `tabs` example originally had the option to add and remove tabs, but this feature was removed due to this issue. 

#### Directional Navigation 

This focus management solution is only a partial one. It allows tabbing between focusable widgets, but this is more common in general purpose applications. For games, it's often much more common to navigate UI directionally, through arrow keys, WASD, gamepads, or other auxiliary devices.

So one possible improvement would be to add this functionality in some shape or form. Maybe as core functionality that extends `FocusTree`. Or maybe as an optional feature or plugin.

#### Focus Groups

Another possible improvement could be adding support for "focus groups." In other words, designating a widget as a container for descendant focusable widgets, but not itself focusable. This could help create "checkpoints" that allow focus to start from the beginning of the group rather than the root widget.

An example would be a complex UI that contains lots of focusable widgets split up between different UI panes. Ideally navigation should start wherever the user currently is. If they clicked on a portion of the UI that contains a focus group, navigation should start there (and can of course cycle out beyond the containing focus group and into others).

This would also be helpful used by the root widget itself, which is currently always focusable. Designating the root node as a focus group would allow cycling through the navigation without including itself in the cycle (which can give the impression that focus has been "lost").


examples/tabs/tab.rs
```diff
@@ -0,0 +1,111 @@
...
+    });
+
+    rsx! {
+        <Background focusable={Some(true)} on_event={Some(event_handler)} styles={Some(border_styles)}>
+            <Background styles={Some(bg_styles)}>
+                <Text content={content} size={12.0} styles={Some(text_styles)} />
+            </Background>
+        </Background>
+    }
+}
```
StarArawn, line 111: Can we make sure we add new lines at the end of files. Its not a huge deal, but the symbol bothers me. lol

examples/tabs/tabs.rs
```diff
@@ -0,0 +1,150 @@
...
+    let text_style = Style {
+        width: StyleProp::Value(Units::Percentage(75.0)),
+        top: StyleProp::Value(Units::Stretch(0.5)),
+        left: StyleProp::Value(Units::Stretch(1.0)),
+        bottom: StyleProp::Value(Units::Stretch(1.0)),
+        right: StyleProp::Value(Units::Stretch(1.0)),
+        ..Default::default()
+    };
+
+    // This is not the most ideal way to generate tabs. For one, the `content` has no access to its actual context
```
StarArawn, line 43: Can we label something with `TODO:` here? I find that helpful at times.

StarArawn (commented):
This seems quite nice! I tested it out and it works well! Do you think it would be difficult to add focus states for text boxes? I think we can address that in a separate PR though! 👍 Just some small cleanup stuff and I think we can merge this!

MrGVSV:
> This seems quite nice! I tested it out and it works well! Do you think it would be difficult to add focus states for text boxes? I think we can address that in a separate PR though! 👍 Just some small cleanup stuff and I think we can merge this!

I actually did have one throughout my testing. I chose to leave it out because I figured that might be something the user wants to configure. But yeah I'll do that in a separate PR so we can look at the options and make sure it's configurable.

Are you thinking a border? Or background color change?

examples/tabs/tab.rs
MrGVSV: Yes! Sorry, I'll do that.

Maybe I'll take care of #41 after this so things are kept consistent.

StarArawn:
> Are you thinking a border? Or background color change?

Whatever looks nicest :) I guess I should update code to add borders..

StarArawn (approved):

PR #51, opened 2022-01-07
Update bevy

StarArawn:


IceSentry (approved):
I tried a couple of examples and everything seemed to work correctly

kayak_render_macros/src/lib.rs
```diff
@@ -154,8 +154,21 @@ pub fn dyn_partial_eq(_: TokenStream, input: TokenStream) -> TokenStream {
...
+                quote!(#ident)
+            }
+        }
+    } else {
+        quote!(kayak_ui::core)
+    };
+
     let result = quote! {{
-        use kayak_core::{Bound, MutableBound};
+        use #kayak_core::{Bound, MutableBound};
```
MrGVSV, line 171: Ah, didn't know I needed to do this. We'll also need to update `use_effect.rs` as well then:
https://github.com/StarArawn/kayak_ui/blob/13458f0806e9df499968b13db21966fb76f40592/kayak_render_macros/src/use_effect.rs#L58

PR #52, opened 2022-01-08
TextBox text style improvements

IceSentry:
* Use colour style for the text
* Use gray colour for the placeholder text
* Add more color constants

I'm not sure if it works on main because it doesn't compile for me, but it did work on the update-bevy branch.

Here's how it looks
image

### DIFF
--- a/examples/text_box.rs
+++ b/examples/text_box.rs
@@ -28,7 +28,7 @@ fn TextBoxExample(context: &mut KayakContext) {
     };
 
     let red_text_styles = Style {
-        color: StyleProp::Value(Color::RED),
+        color: StyleProp::Value(Color::new(1., 0., 0., 1.)),
         ..input_styles.clone()
     };
 
--- a/examples/text_box.rs
+++ b/examples/text_box.rs
@@ -4,6 +4,7 @@ use bevy::{
     DefaultPlugins,
 };
 use kayak_core::Color;
+use kayak_render_macros::use_state;
 use kayak_ui::bevy::{BevyContext, BevyKayakUIPlugin, FontMapping, UICameraBundle};
 use kayak_ui::core::{
     render, rsx,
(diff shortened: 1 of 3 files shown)


kayak_core/src/color.rs
```diff
@@ -34,6 +34,38 @@ impl Color {
...
+    };
+
+    /// The blue color.
+    pub const BLUE: Color = Color {
+        r: 0.0,
+        g: 0.0,
+        b: 1.0,
+        a: 1.0,
+    };
+
```
MrGVSV, line 68: This might need to change due to your other PR (#53)

IceSentry: Yeah, I think I'll just hardcode the red in this PR and once the color is merged I'll update this and any other place that uses hardcoded colors

examples/text_box.rs
```diff
@@ -13,35 +15,43 @@ use kayak_ui::widgets::{App, OnChange, TextBox, Window};
 
 #[widget]
 fn TextBoxExample(context: &mut KayakContext) {
-    let value = context
-        .create_state("I started with a value!".to_string())
-        .unwrap();
-    let value2 = context.create_state("".to_string()).unwrap();
+    let (value, set_value, _) = use_state!("I started with a value!".to_string());
+    let (empty_value, set_empty_value, _) = use_state!("".to_string());
+    let (red_value, set_red_value, _) = use_state!("This text is red".to_string());
```
MrGVSV, line 20: With this change could you remove the `Bound` and `MutableBound` imports? They're automatically added by the `use_state!` macro.

IceSentry: Yep, sorry, rust-analyzer is going a bit crazy with me constantly changing branches and recompiling

MrGVSV: No worries!

MrGVSV (approved):

StarArawn (approved):

PR #53, opened 2022-01-08
add primary colors

IceSentry:
I also added a new rgb constructor and made the existing constructor `const`. 

I wonder if it would be possible to simply reuse the bevy Color type since it's much more complete.

### DIFF
--- a/kayak_core/src/color.rs
+++ b/kayak_core/src/color.rs
@@ -19,30 +19,27 @@ impl Default for Color {
 
 impl Color {
     /// The black color.
-    pub const BLACK: Color = Color {
-        r: 0.0,
-        g: 0.0,
-        b: 0.0,
-        a: 1.0,
-    };
+    pub const BLACK: Color = Color::rgb(0.0, 0.0, 0.0);
 
     /// The white color.
-    pub const WHITE: Color = Color {
-        r: 1.0,
-        g: 1.0,
-        b: 1.0,
-        a: 1.0,
-    };
+    pub const WHITE: Color = Color::rgb(1.0, 1.0, 1.0);
+
+    pub const GRAY: Color = Color::rgb(0.5, 0.5, 0.5);
+    pub const RED: Color = Color::rgb(1.0, 0.0, 0.0);
+    pub const GREEN: Color = Color::rgb(0.0, 1.0, 0.0);
+    pub const BLUE: Color = Color::rgb(0.0, 0.0, 1.0);
+    pub const YELLOW: Color = Color::rgb(1.0, 1.0, 0.0);
+    pub const MAGENTA: Color = Color::rgb(1.0, 0.0, 1.0);
+    pub const CYAN: Color = Color::rgb(0.0, 1.0, 1.0);
 
     /// A color with no opacity.
-    pub const TRANSPARENT: Color = Color {
-        r: 0.0,
-        g: 0.0,
-        b: 0.0,
-        a: 0.0,
-    };
+    pub const TRANSPARENT: Color = Color::new(0.0, 0.0, 0.0, 0.0);
 
-    pub fn new(r: f32, g: f32, b: f32, a: f32) -> Self {
+    pub const fn new(r: f32, g: f32, b: f32, a: f32) -> Self {
         Self { r, g, b, a }
+    }
+
+    pub const fn rgb(r: f32, g: f32, b: f32) -> Self {
+        Self { r, g, b, a: 1.0 }
     }
 }


MrGVSV:
> I wonder if it would be possible to simply reuse the bevy Color type since it's much more complete.

I think we'd want to avoid a strict Bevy dependency in the core crate (unless we decide to be a purely Bevy project).

IceSentry:
Yeah, I agree. I was more thinking about if it would be possible to just copy it.

kayak_core/src/color.rs
```diff
@@ -19,30 +19,27 @@ impl Default for Color {
...
-    pub const TRANSPARENT: Color = Color {
-        r: 0.0,
-        g: 0.0,
-        b: 0.0,
-        a: 0.0,
-    };
+    pub const TRANSPARENT: Color = Color::new(0.0, 0.0, 0.0, 0.0);
 
-    pub fn new(r: f32, g: f32, b: f32, a: f32) -> Self {
+    pub const fn new(r: f32, g: f32, b: f32, a: f32) -> Self {
```
MrGVSV, line 38: Would it be worth renaming this to `rgba` for clarity?

MrGVSV (approved):

MrGVSV:
> Yeah, I agree. I was more thinking about if it would be possible to just copy it.

We could definitely do something like that. But I’m wondering if we want all their colors or if we want to pick and choose our own 🤔

kayak_core/src/color.rs
IceSentry: Yes, but I can't build this on main, I can only build on the update-bevy branch and I didn't want to do a refactor that would affect other files.

IceSentry:
I'm fine with having only primary colors as default constants, but I think having access to multiple color space constructor and all the other operations defined on the bevy Color type would be nice to have

kayak_core/src/color.rs
MrGVSV: Oh strange. Is it a Kayak issue or your local system? Have you tried cleaning and rebuilding?

IceSentry: I'm not sure honestly. I tried `cargo clean && rm Cargo.lock` many times and then building with both nightly and stable. I always get 
```bash
   Compiling bevy_render v0.5.0 (https://github.com/bevyengine/bevy?rev=081350916c39fc1c7c9d659ccb6d86413832b465#08135091)
error: cannot find derive macro `Component` in this scope
  --> C:\Users\<username>\.cargo\git\checkouts\bevy-f7ffde730c324c74\0813509\crates\bevy_render\src\render_resource\texture.rs:12:24
   |
12 | #[derive(Clone, Debug, Component)]
   |                        ^^^^^^^^^
   |
   = note: consider importing one of these items:
           bevy_ecs::prelude::Component
           crate::Component

error: could not compile `bevy_render` due to previous error
```

IceSentry: But with the update-bevy branch I can build as long as I use stable

MrGVSV: Do you also need to remove `C:\Users\<username>\.cargo\git\checkouts\bevy-f7ffde730c324c74/...`?

It seems to build fine on my end.

IceSentry: So, I deleted the all the `.cargo\git\checkouts\bevy-*` and it works now. I genuinely don't understand what happened. It's clearly a build artifact from cargo, why would it use the wrong one though.

MrGVSV: I wish I knew 😅

IceSentry: So, now that I can compile I did the change and it still compiles

MrGVSV (approved):

kayak_core/src/color.rs
```diff
@@ -19,30 +19,27 @@ impl Default for Color {
 
 impl Color {
     /// The black color.
-    pub const BLACK: Color = Color {
-        r: 0.0,
-        g: 0.0,
-        b: 0.0,
-        a: 1.0,
-    };
+    pub const BLACK: Color = Color::rgb(0.0, 0.0, 0.0);
```
StarArawn, line 22: We can also do something like:
```rust
#[cfg(not(feature = "bevy_renderer"))]
pub use color::Color;
#[cfg(feature = "bevy_renderer")]
pub use bevy::prelude::Color;
```

StarArawn (changes requested):

kayak_core/src/color.rs
MrGVSV: This would force us to ensure we have a similar API to bevy anyways , though, right? Otherwise in all the other crates and for the default widgets, we'll have to add these feature lines as well (not terrible but not super ideal). 

I’d prefer adding `From<Color>` impls for both under the bevy feature. Then the user can choose to use either at any time and we don't have to feature-flag internal color-related code.

kayak_core/src/color.rs
```diff
@@ -19,30 +19,27 @@ impl Default for Color {
 
 impl Color {
```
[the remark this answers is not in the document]
MrGVSV: We could, and that's what @IceSentry suggested as well. My hesitance with it was that we have an opportunity to do things in ways bevy doesn't. Their code is likely good enough, but maybe there'a a better library or method we could use that handles more or handles things better.

I don't know, that was just my reasoning at least 😅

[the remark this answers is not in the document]
IceSentry: I'm just not sure how it works from a license point of view. If I just copy it all, should I add a comment pointing to the bevy Color impl?

kayak_core/src/color.rs
IceSentry: I like the `From<Color>` idea, but I'm not familiar enough with rust to do that. I'm not exactly sure how that would look like. Would I just need to impl the From trait on color::Color and that's it?

MrGVSV: No problem! Yep, something like:

```rust
#[cfg(feature = "bevy_renderer")]
impl From<bevy::prelude::Color> for crate::Color { /* ... */ }
```

And then the same for the reverse.

kayak_core/src/color.rs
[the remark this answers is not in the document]
MrGVSV: I'm not sure either to be honest. We've done it with other things (`KeyCode`, for example), but not something that contains a bunch of unique logic. I think it could fall under "re-distribution", meaning we might need to add a license? (For MIT at least)

kayak_core/src/color.rs
StarArawn: Lets go with `From<Color>` :)

kayak_core/src/color.rs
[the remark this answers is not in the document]
StarArawn: > I'm not sure either to be honest. We've done it with other things (`KeyCode`, for example), but not something that contains a bunch of unique logic. I think it could fall under "re-distribution", meaning we might need to add a license? (For MIT at least)

I think a lot of the keycode stuff was copied from winit as well, but it would probably be a good idea to include the bevy license somewhere. Perhaps as a link at the top of the file?

PR #54, opened 2022-01-08
Separated `KeyboardInput` into `KeyUp` and `KeyDown` events

MrGVSV:
Split the `KeyboardInput` event into separate `KeyUp` and `KeyDown` events. Before, we were receiving two events for every press/release of a key when we should only be receiving one.

~~One possible oddity, is that the `KeyDown` event fires for every repeat (as long as the key is held down).~~ Apparently this [also happens for HTML](https://jsfiddle.net/zdgaj8mx/), so maybe this is fine?

Additionally, added a `KeyboardEvent` struct that allows us to pass along metadata with these new events (they replaced the single `key` variant member). It stores a `KeyboardModifiers` struct that contains which modifier keys are actively being pressed (something we can't know using the old system— not easily at least).

Again, I used W3's [specifications](https://www.w3.org/TR/uievents/#events-keyboard-types) for reference, but we don't strictly need to follow those to a T. So let me know if there's a different way we should be going about things here!

### DIFF
--- a/kayak_core/src/event_dispatcher.rs
+++ b/kayak_core/src/event_dispatcher.rs
@@ -1,5 +1,5 @@
 use std::collections::{HashMap, HashSet};
-use crate::{Event, EventType, Index, InputEvent, InputEventCategory, KayakContext, PointerEvents};
+use crate::{Event, EventType, Index, InputEvent, InputEventCategory, KayakContext, KeyboardEvent, KeyboardModifiers, KeyCode, PointerEvents};
 use crate::layout_cache::Rect;
 use crate::widget_manager::WidgetManager;
 
@@ -35,6 +35,7 @@ pub(crate) struct EventDispatcher {
     current_mouse_position: (f32, f32),
     next_mouse_position: (f32, f32),
     previous_events: EventMap,
+    keyboard_modifiers: KeyboardModifiers,
 }
 
 impl EventDispatcher {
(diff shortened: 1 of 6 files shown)


kayak_core/src/event_dispatcher.rs
```diff
@@ -303,9 +304,27 @@ impl EventDispatcher {
...
-                InputEvent::Keyboard { key } => event_stream.push(
-                    Event::new(current_focus, EventType::KeyboardInput { key: *key })
-                ),
+                InputEvent::Keyboard { key, is_pressed } => {
+                    // === Modifers === //
+                    match key {
+                        KeyCode::LControl | KeyCode::RControl => self.keyboard_modifiers.is_ctrl_pressed = *is_pressed,
+                        KeyCode::LShift | KeyCode::RShift => self.keyboard_modifiers.is_shift_pressed = *is_pressed,
+                        KeyCode::LAlt | KeyCode::RAlt => self.keyboard_modifiers.is_alt_pressed = *is_pressed,
+                        KeyCode::LWin | KeyCode::RWin => self.keyboard_modifiers.is_meta_pressed = *is_pressed,
```
StarArawn, line 313: Do you know what macos uses for command? LWin and RWin are the same as command for macos?

MrGVSV: Yeah it's the same for both.

I tried to suggest renaming it in an [issue](https://github.com/bevyengine/bevy/issues/3240) but some people didn't like the term "meta".

I think it's fine here, though, since we're trying to be sorta like React and the web uses that [terminology](https://www.w3.org/TR/uievents/#dom-keyboardevent-metakey). But let me know if you prefer a different term!

MrGVSV: Actually if we use "meta" or some other term should we also update our KeyCode enum to match? Or do we remain consistent with Bevy? 🤔

StarArawn: I think if we keep LWin and RWin it's fine. I don't want to diverge too much from bevy.

MrGVSV: Do you want to also change `is_meta_pressed` to `is_win_pressed`? Or are we just keeping the keycodes themselves consistent with bevy/winit?

PR #56, opened 2022-01-10
[RFC 1] Widget Restructuring

MrGVSV:
This RFC proposes a possible restructuring of widgets to allow for better separation of logic and a (hopefully) improved API.

### DIFF
--- a/rfcs/widget-restructure-rfc-1.md
+++ b/rfcs/widget-restructure-rfc-1.md
@@ -428,6 +428,12 @@ impl Widget for MyWidget {
 
 ---
 
+## Tl;DR
+
+1. Add `KayakContextRef` and update `Widget` methods - to interface with  `KayakContext` safely
+2. Add `WidgetConstructor` and `BaseProps` - to consistently generate widgets in macros
+3. Generate Widget Context during `rsx` expansion - to interface with `KayakContext` and the widget safely
+
 ## Issues
 
 ### Protection
--- a/rfcs/widget-restructure-rfc-1.md
+++ b/rfcs/widget-restructure-rfc-1.md
@@ -145,6 +145,7 @@ The derive macro would have some associated macros, which will be used to mark c
 * `#[props(OnEvent)]` - Used to mark a field as the `OnEvent` prop
 * `#[props(Styles)]` - Used to mark a field as the `Styles` prop
 * `#[props(Focusable)]` - Used to mark a field as the `Focusable` prop
+* `#[props(Children)]` - Used to mark a field as the `Children` prop
 
 There may be more added to this list in the future, but for now this is the main assortment.
 
@@ -161,6 +162,8 @@ struct MyWidgetProps {
   event_handler: OnEvent
   #[props(Focusable)]
   focusable: Optional<bool>,
+  #[props(Children)]
+  children: Children,
   // Defined without the marker attribute:
   styles: Styles,
 }
@@ -179,6 +182,11 @@ impl WidgetProps {
     self.focusable
   }
   
+  // Children is an alias for an Option so it doesn't need to return Option<Children>
+  fn get_children(&self) -> Children {
+    self.children.clone()
+  }
+  
   // ...
 }
 ```
@@ -188,7 +196,7 @@ impl WidgetProps {
 Another addition to `Widget` would be the constructor method. This associated method will allow widgets to be generated with a set of props in a more controlled and freeform way. The method looks like this:
 
 ```rust
-fn constructor(id: Index, children: Children, mut props: Self::Props) -> dyn Widget<Props=Self::Props> where Self: Sized;
+fn constructor(mut props: Self::Props) -> dyn Widget<Props=Self::Props> where Self: Sized;
 ```
 
 > Note that this would put the `Sized` restriction on widgets.
@@ -209,7 +217,6 @@ To define a custom widget, we need to first define our struct:
 #[derive(Debug)]
 struct MyButton {
   my_id: Index,
-  my_children: Children,
   my_props: MyButtonProps,
 }
 ```
@@ -360,7 +367,7 @@ impl<'a> KayakContextRef<'a> {
     self.widget_tree.add(child_id, Some(id));
     if should_rerender {
         let mut child_widget = self.context.widget_manager.take(child_id);
-        let mut context = KayakContextRef { context: self.context, current_id: id };
+        let mut context = KayakContextRef { context: self.context, current_id: child_id };
         child_widget.render(context);
         self.context.widget_manager.repossess(child_widget);
     }
@@ -456,6 +463,12 @@ In order to properly get the props struct, we can instead define our widget like
 ```rust
 #[widget]
 fn MyWidget(props: MyWidgetProps) {
+  // Internally insert:
+  // ```
+  // let props = self.props.clone();
+  // ```
+  
+  
   let WidgetProps {foo, bar} = props;
   // ...
 }
--- a/rfcs/widget-restructure-rfc-1.md
+++ b/rfcs/widget-restructure-rfc-1.md
@@ -15,19 +15,29 @@ So in summary the two main issues are:
 1. Internal widget data is exposed to the user in the render function in a confusing manner
 2. Error-prone `KayakContext` methods are also exposed to the user in the render function when they shouldn't be
 
-## Solution *(Version 2)*
+## Solution *(Version 3)*
 
 > For Version 1, check out this commit: [841ff97fa68](https://github.com/StarArawn/kayak_ui/pull/56/commits/841ff97fa68b0e3919a42f871034383de10b2f68)
+>
+> For Version 2, check out these commits: [841ff97fa68..9a246ee1a](https://github.com/StarArawn/kayak_ui/pull/56/files/841ff97fa68b0e3919a42f871034383de10b2f68..9a246ee1a8fbc1e543f6ae58d5a18416f9e20dbf)
 
-### Issues and Differences with Version 1
+### Recap
 
-One issue with the solution proposed in Version 1 of this RFC, is that it only focused on those choosing to use `rsx`. Custom widgets would still need to go through a lot of boilerplate code in order to get up and running. Additionally, these custom widgets needed to match the schema of their `rsx` counterparts, down to the name (more-or-less).
+#### Version 1
 
-This updated version attempts to target these users as well, since customizing a widget to your liking is definitely a great thing to have. I'll go through each section in order of increasing automation (from manually writing code to using mostly macros). But first, changes to `Widget` and how it renders.
+This version proposed changes that would add a layer of abstraction between a functional component, its generated widget struct, and the `KayakContext`. This was good for people writing functional components, but offered nothing for those defining their widgets manually.
 
----
+It also added the concept of placing all user-defined props in a generated `<Widget Name>Props` struct to help isolate user-defined data with Kayak-defined data.
 
-###  Widget
+#### Version 2
+
+Version 2 sought to take the concepts of Version 1 and apply some of them across manually defined widgets as well. To do this, the `KayakContextRef` was created, which acts as a temporary interface to `KayakContext` and is what would be passed to `Widget` implementors, instead of the raw context. Functional widgets would continue to use the stricter `<Widget Name>Context` to interface.
+
+It also suggested bringing a props struct to manual implementors as well. To do this nicely it proposed a `WidgetConstructor` trait, which defined an associated `Props` type. After implementing, a widget could be created by calling `MyWidget::construct(id, children, props)`. And `props` would be generated like this `MyWidget::create_props(base_props)`.
+
+This worked but created a disparity between widgets: a widget could choose to not impl `WidgetConstructor`, preventing it from being composed in tags (i.e., `<MyWidget />`). It also forced the `Props` type to return a default struct (not necessarily by implementing `Default`), which isn't well-suited for required props.
+
+### Widget
 
 Previously we had the `WidgetContext` act as the barrier between the render method and the widget/context. This works well for `rsx`-generated widgets, but not so much for manually defined widgets. The reason is that by using `WidgetContext` we place a lot more expectations on how the rendering should be done. Again, one of the goals with this RFC is to cut back on expectations placed on the user.
 
@@ -35,15 +45,22 @@ And while `WidgetContext` isn't ideal for these scenarios, it still provides a f
 
 #### KayakContextRef
 
-This struct will be the interface into `KayakContext`. In it, we can place code specifically for widgets to use during their render phase.
+The first issue to address is providing safe access to `KayakContext`. By "safe," I'm referring to the fact that currently access to `KayakContext` is unfiltered. This allows the user to run code they shouldn't or come up with unmaintainable hacks. To reduce this, we can add a struct that is temporarily created by `KayakContext` to act as a safer interface to itself. 
+
+This is where `KayakContextRef` comes in:
 
 ```rust
 struct KayakContextRef<'a> {
   context: &'a mut KayakContext,
   current_id: Index,
+  // other fields...
 }
 
 impl<'a> KayakContextRef<'a> {
+  pub(crate) fn new(context: &mut KayakContext, current_id: Index) {
+    // ...
+  }
+  
   /// Expose the bind method to the widget
   pub fn bind(...) {
     /// Forward the call
@@ -69,7 +86,7 @@ This changes what `Widget::render` looks like:
 ```rust
 trait Widget {
   // ...
-  fn render(&mut self, mut context: KayakContextRef);
+  fn render(&mut self, context: &mut KayakContextRef);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,385 @@
...
+}
+
+impl Widget for MyWidget {
+  fn render(&mut self, context: &mut KayakContext) {
+    Self::render(self.props, context);
+  }
+}
+
+impl MyWidget {
+  fn render(props: MyWidgetProps, context: &mut KayakContext) {
```
StarArawn, line 59: I like this, but where do the props come from into the `render`?

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,385 @@
...
+### Widget Context
+
+Now comes the second issue of direct access to `KayakContext`. I think a good way of solving this would be to generate an interface struct. This struct would act as a middleman between the render function and `KayakContext`, forwarding only select methods to the context.
+
+There are two ways of going about this.
+
+1. Create a local struct:
+
+```rust
+struct MyWidgetContext<'a> {
```
StarArawn, line 74: I'm worried anytime I see: `<'a>` because it can mean that the lifetime trickles up and we end up with lifetime requirements that are annoying for end users to deal with.

StarArawn: Does this context get created on render and then dropped after?

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Problem
+
+A lot of internal stuff is exposed to the end user that either doesn't need to be or shouldn't be. At best it can be confusing or misleading to users, and at worst it can cause some users to use code in ways they shouldn't (i.e., unintentional hacks that are unstable and could break between versions).
+
+For example, it seems wrong that a user can do this: `self.set_id(Index::default())`. And while this might not be that big of an issue ( "users should just *not* do that" ), what if we want to add other functionality onto widgets in the future? Maybe we add the ability to give widgets their own classes for theming. This might be confusing for the user who thinks that `self.add_class("foo")` should work on its own (but might not since the context also needs to  be made aware of the change).
+
+Additionally, `KayakContext`, has a lot of functionality that doesn't need to be exposed within the context of a widget, but needs to elsewhere (such as the `KayakContext::render` and `KayakContext::process_events` methods). Methods like these need to be public, but they're dangerous to use within a widget's render method.
```
StarArawn, line 11: This is something that has been bothering me about the API since I've started this. I think this makes sense to clean this up and prevent users from doing things that shouldn't be allowed! 👍

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,385 @@
...
+use crate::render_command::RenderCommand;
+use crate::styles::{Style, StyleProp};
+
+mod my_widget_module {
+    use derivative::Derivative;
+    use crate::{Color, Index, KayakContext, OnEvent, Widget};
+    use crate::styles::Style;
+
+    #[derive(Derivative)]
+    #[derivative(Default, Debug, PartialEq)]
```
StarArawn, line 232: Do we still need `PartialEq` here? 🤔

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,385 @@
...
+    pub struct MyWidget {
+        pub id: Index,
+        pub props: MyWidgetProps,
+        pub styles: Option<Style>,
```
StarArawn, line 236: I'm wondering if styles should be included in `MyWidgetProps` instead?

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,385 @@
...
+            let this = self.generate_context(context);
+            if let Some(on_event) = this.widget.on_event.as_mut() {
+                if let Ok(mut on_event) = on_event.0.write() {
+                    on_event(this.context, event);
+                }
+            }
+        }
+
+        fn render(&mut self, context: &mut KayakContext) {
+            Self::render(self.generate_context(context));
```
StarArawn, line 309: Isn't this a recursive loop?

StarArawn: Might be better to say:
```suggestion
            MyWidget::render(self.generate_context(context));
```

StarArawn (commented):
Just some initial thoughts. I need to read through it some more and think careful about some things.

rfcs/widget-restructure-rfc-1.md
MrGVSV: This should probably be called `render_internal` or something.

Also, this is only part of the proposed changes (I was trying to show my thought process leading up to the final proposal). In the end, this would actually look more like:

```rust
impl Widget for MyWidget {
  // ...
  fn render(&mut self, context: &mut KayakContext) {
    let mut widget_context = /* ... create MyWidgetContext ... */
    // Pass off to associated method so it loses access to dangerous MyWidget internals
    Self::render_internal(widget_context);
  }
}

impl MyWidget {
  fn render_internal(mut widget_context: MyWidgetContext) {
    // Render logic from functional component
  }
}
```

So it would be called by the `MyWidget` instance, specifically in its implementation of the `Widget::render` trait method:

`KayakContext` -(via `Box<dyn Widget>`)-> `MyWidget::render(&mut self, ...)` -> `MyWidget::render_internal()`

rfcs/widget-restructure-rfc-1.md
MrGVSV: > I'm worried anytime I see: `<'a>` because it can mean that the lifetime trickles up and we end up with lifetime requirements that are annoying for end users to deal with.

Yeah, but I'm not sure how else to split up the internal widget code from the user-facing API.

> Does this context get created on render and then dropped after?

Yeah, this is a temporary type that basically serves as an ephemeral interface to the widget and context. And can be used to track changes.

rfcs/widget-restructure-rfc-1.md
MrGVSV: I'm not sure. I basically copied it from a macro expansion on the `TextBox` widget.

rfcs/widget-restructure-rfc-1.md
MrGVSV: Yeah, that's an option (pun not intended, but happily embraced)!

I think the only big reason to have an outer `styles` field would be to check if there were any changes to it in the actual render. But we likely don't need something like that, so it should be fine in `props`.

rfcs/widget-restructure-rfc-1.md
MrGVSV: > Isn't this a recursive loop?

No I meant to call this something else. Maybe `render_internal`?

It's an associated method being called from an instance method, so it shouldn't recurse.

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,385 @@
...
+Then if we need to handle any changes we can just iterate through the set of changes and handle them individually.
+
+```rust
+impl Widget for MyWidget {
+  // ...
+  fn on_event(&mut self, context: &mut KayakContext, event: &mut crate::Event) {
+    let mut this = self.generate_context(context);
+    if let Some(mut on_event) = this.widget.on_event.clone() {
+      if let Ok(mut on_event) = on_event.0.write() {
+        on_event(this, event);
```
MrGVSV, line 375: Actually this should be:

```suggestion
        on_event(&mut this, event);
```

Since we want to read from it after passing the event.

rfcs/widget-restructure-rfc-1.md
StarArawn: I think this makes sense. How would this work for people who want to not use `rsx!` though? 🤔

rfcs/widget-restructure-rfc-1.md
StarArawn: I think the lifetime is fine here! 👍 Since it's only created and dropped we are not storing it anywhere we should be fine.

rfcs/widget-restructure-rfc-1.md
StarArawn: Yeah `render_internal` sounds fine.

rfcs/widget-restructure-rfc-1.md
MrGVSV: My thought was that this is all for users who want to write functional components (i.e. they don't want to write their own super-custom logic or boilerplate). And for users who prefer to write it all out manually, then they'll have to write their logic on their own (which isn't a bad thing if the point is they want more control over functionality).

And since `render_internal`, `MyWidgetContext`, etc. are all generated by the macro, they don't need to implement or recreate any of it. They can do pretty much what we have now for [`Fragment`](https://github.com/StarArawn/kayak_ui/blob/dfdee36240be1236081c480cb8f573ab6c43e586/kayak_core/src/fragment.rs) and [`VecTracker`](https://github.com/StarArawn/kayak_ui/blob/dfdee36240be1236081c480cb8f573ab6c43e586/kayak_core/src/vec.rs).

So if they want to "add a class for theming", they'd have to do something like:

```rust
impl MyCustomWidget {
  fn add_class(&mut self, context: &mut KayakContext, class_name: &str) {
    println!("Added the class: {}", class_name);
    self.class_list.insert(class_name.to_string());
    context.add_class(self.id, class_name);
    // Other custom stuff...
  }
}

impl Widget for MyCustomWidget {
  fn render(&mut self, context: &mut KayakContext) {
    // Do custom render logic (as we do now basically)
    self.add_class(context, "foo");
    // ...
  }
}
```

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,554 @@
...
+
+```rust
+// fragment.rs
+
+fn render(&mut self, mut context: KayakContextRef) {
+	let tree = context.create_tree();
+  if let Some(children) = self.children.take() {
+    context.add_children(children, &tree);
+  } else {
+      return;
```
StarArawn, line 226: I think we can always call commit tree here. If no changes were made than merge doesn't do anything.

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,554 @@
...
+  }
+  context.commit_tree(tree);
```
StarArawn, line 228: I think we can just call this `commit` perhaps?

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,554 @@
...
+
+```rust
+// <MyWidget foo={Some(1)} bar={"Hello".to_string()} />
+let mut props = MyWidget::create_props(base_props);
+props.foo = Some(1);
+props.bar = "Hello".to_string();
+let widget = MyWidget::construct(widget_id, children, props);
+```
+
+As for the implementation of `Widget:;render`, it will look largely the same as the manual widgets. The main difference is that now props are no longer found directly on the struct.
```
StarArawn, line 317:
```suggestion
As for the implementation of `Widget::render`, it will look largely the same as the manual widgets. The main difference is that now props are no longer found directly on the struct.
```

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,554 @@
...
+}
+
+#[derive(Default)]
+struct MyProps {
+  foo: Option<i32>,
+  bar: String
+}
+```
+
+A derive would be nice since in order to reduce boilerplate, but it's not necessary. And this may add an extra layer of confusion (due to required marker attributes and the `Default` bound on the props). But I left it here as a possible option and to see if there are other, better ways of implementing something like this.
```
StarArawn, line 342: I like making life easier :D

StarArawn (changes requested):

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,554 @@
...
+}
+```
+
+And results in usage like so:
+
+```rust
+// fragment.rs
+
+fn render(&mut self, mut context: KayakContextRef) {
+	let tree = context.create_tree();
```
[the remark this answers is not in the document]
MrGVSV: That's a much cleaner idea. Let's do that!

rfcs/widget-restructure-rfc-1.md
MrGVSV: I named it to be a bit more clear amongst its other methods (`bind`, `create_state`, etc.). But we can go ahead and just call it `commit` I guess since it only takes a tree anyways haha

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,554 @@
...
+<br />
+
+#### Constructed Widgets
+
+A constructed widget, for the purposes of this RFC, is one that is manually defined like above but uses some traits and derives to allow it to be used as a tagged (`</>`) component. Like the manually defined widgets, it's good for adding customization and extending the widget itself with other functionality.
+
+To integrate with the `rsx` tagging syntax without becoming a fully "functional widget," it must implement `WidgetConstructor` and define its props.
+
+```rust
+trait WidgetConstructor {
```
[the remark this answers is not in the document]
MrGVSV: I'd prefer this, but we'd have to specify the `Props` type on `WidgetConstructor` I think:

```rust
trait Widget: WidgetBase + WidgetConstructor<Props=MyProps>;
```

Which would bubble up to `Widget`, preventing us from using it as a trait object if I’m correct.

rfcs/widget-restructure-rfc-1.md
MrGVSV: Lol my keyboard is a bit broken. You may see double letters sometimes as well 😅

rfcs/widget-restructure-rfc-1.md
MrGVSV: Yeah! I think this either needs to be cleaned up or just super well-documented for it to actually be easier haha

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,667 @@
...
+
+A custom widget is a widget defined manually by a user using standard Rust syntax (i.e. not by using the `#[widget]` attribute). By defining a widget manually, a user has finer control over behavior, data, and more.
+
+To define a custom widget, we need to first define our struct:
+
+```rust
+#[derive(Debug)]
+struct MyButton {
+  my_id: Index,
+  my_children: Children,
```
StarArawn, line 212: I think this can move into the props as well unless there is a specific reason we don't want to?

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,667 @@
...
+
+Now with everything defined, we're ready to implement `Widget`.
+
+Firstly, the constructor:
+
+```rust
+impl Widget for MyButton {
+  type Props = MyButtonProps;
+  
+  fn constructor(id: Index, children: Children, mut props: Self::Props) -> dyn Widget<Props=Self::Props> where Self: Sized {
```
StarArawn, line 256: I'm not sure we actually have an id at this point. We normally get an ID when this is called:
`context.widget_manager.create_widget(0usize, built_widget, parent_id);`

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,667 @@
...
+    }
+  }
+  
+  // Pretend `Children` isn't wrapped by `Option<...>`
+  pub fn add_children(&mut self, children: Children) {
+    let id =  self.current_id;
+    children(self.widget_tree.clone(), Some(id), context);
+  }
+
+  pub fn add_widget<W: Widget>(&mut self, widget: W, widget_index: usize) {
```
StarArawn, line 355: ID here should be the `parent_id` I think in this case its the `self.current_id`, but on a second call to `add_widget` `current_id` might be the `child_widget`'s id. Might be worth resetting `current_id` at the end of this function.

rfcs/widget-restructure-rfc-1.md
```diff
@@ -0,0 +1,667 @@
...
+  pub props: MyWidgetProps,
+}
+```
+
+In order to properly get the props struct, we can instead define our widget like so:
+
+```rust
+#[widget]
+fn MyWidget(props: MyWidgetProps) {
+  let WidgetProps {foo, bar} = props;
```
StarArawn, line 459: Internally we should probably `Clone` props so we have values at this point. I think this code would resolve to:
```rust
fn render(...) {
  let props = self.props.clone();
  let WidgetProps { foo, bar } = props;
  
  ...
}
```

StarArawn (commented):
Just a few very small notes. :) Overall this is looking great! 👍

rfcs/widget-restructure-rfc-1.md
MrGVSV: Nope. I'll do that!

rfcs/widget-restructure-rfc-1.md
MrGVSV: Gotcha, so should it be set to `Index::default()`?

rfcs/widget-restructure-rfc-1.md
MrGVSV: So update to widget's ID if `should_rerender` (before rendering the widget)? Then reset back to the original at the end of the method?

rfcs/widget-restructure-rfc-1.md
StarArawn: Yeah, that sounds right!

rfcs/widget-restructure-rfc-1.md
MrGVSV: Actually it looks like I have it setup to create a new `KayakContextRef` before rendering the child (although I forgot to give it `child_id`). So it shouldn't need to be updated, right?

StarArawn: Hmm, yeah that should be fine 👍

PR #58, opened 2022-01-11
[RFC 2] Text Measure

StarArawn:


rfcs/rfc-2-font-measuring.md
```diff
@@ -0,0 +1,58 @@
...
+let mut font = KayakFont::new(
+    Sdf::from_bytes(include_bytes!("./assets/roboto.kayak_font")),
+    atlas_handle,
+);
+```
+
+## Additional issues:
+The text widget will need to be re-rendered when it's font is loaded in. I suggest quite a simple way of doing this:
+```rust
+fonts: Binding<HashMap<u16, KayakFont>>,
```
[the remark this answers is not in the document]
StarArawn: The main reason for `u16` is that its a default type for rust. If we used UUID people would have to depend on the same UUID crate or we'd have to expose a way of generating UUID's?

rfcs/rfc-2-font-measuring.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Alternatives
+Another solution is to use an additional data structure:
+```rust
+fonts: HashMap<u16, KayakFont>,
+font_bindings: HashMap<u16, Binding<u16>>,
+```
+
+Unresolved issues with this method:
+1. How do we initially populate the binding? For example if font id 1 isn't loaded in yet, but a Text widget asks for 2? Do we create it then?
```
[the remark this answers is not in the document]
StarArawn: Yes that would help perhaps `AssetId` as a name?

rfcs/rfc-2-font-measuring.md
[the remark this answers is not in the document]
MrGVSV: > The main reason for `u16` is that its a default type for rust. If we used UUID people would have to depend on the same UUID crate or we'd have to expose a way of generating UUID's?

That's true. Yeah, we could internalize the UUID and create an interfacing struct. Which is probably the better solution (that's what Bevy did with `HandleId`). And if we want to stick to primitives, we _may_ be able to use a `u128`, but I think a struct would be better.

rfcs/rfc-2-font-measuring.md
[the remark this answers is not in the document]
MrGVSV: > Yes that would help perhaps `AssetId` as a name?

Hm, would we ever want to have something like:

```rust
struct Asset {
  id: AssetId,
  load_state: AssetLoadState,
}
```

My only issue with `AssetId` is that it seems like it should be used as an ID, when really, it contains the ID, which could cause issues in a `HashMap<AssetId, SomeAsset>` (unless we mark `load_state` and other potential fields as ignored for `PartialEq`, `Hash`, etc.).

But I also don't love `Asset` either since it makes it seem like that is the asset itself or somehow contains it, which it doesn't. Maybe `AssetHandle`?

rfcs/rfc-2-font-measuring.md
```diff
@@ -10,49 +10,54 @@ I propose two different solutions for the issues above.
...
+    ..
+    // Stores AssetStorage generically.
+    assets: Resources,
+}
+
+impl KayakContext {
+    pub fn get_asset<T>(&self, asset_handle: AssetHandle<T>) -> &T {
+        // Throw error if AssetStorage doesn't exist in the asset resources.
+        ..
+    }
```
MrGVSV, line 48: I feel like this should return a `Result` instead. Then we can create an asset-related error type that makes handling the error a bit easier.

~~Also, should this return a `&Binding<T>`?~~ Actually, it probably shouldn't since they can't bind to it. What would be the method of attaching/removing a listener for an asset on a given widget?

rfcs/rfc-2-font-measuring.md
```diff
@@ -10,49 +10,54 @@ I propose two different solutions for the issues above.
...
+
+    // We want binding here because we need to allow widgets to track changes.
+    pub fn set_asset<T>(&mut self, asset: T) -> Binding<AssetHandle<T>> {
+        ..
+    }
```
MrGVSV, line 53: Is this called by the widget? How does it get access to `T`? Or is this called in some other context?

StarArawn: `set_asset` would be called by `bevy_kayak_ui`.

rfcs/rfc-2-font-measuring.md
StarArawn: Sorry you are right it probably should be `Binding<T>`. I think I mixed some things up.

MrGVSV: And this method will automatically bind the widget to the `Binding<T>` then?

rfcs/rfc-2-font-measuring.md
MrGVSV: Gotcha. So we return the `Binding` just in case the setter needs it? And the widget uses `get_asset` to get their own binding?

rfcs/rfc-2-font-measuring.md
```diff
@@ -42,13 +31,19 @@ pub struct KayakContext {
 }
 
 impl KayakContext {
-    pub fn get_asset<T>(&self, asset_handle: AssetHandle<T>) -> &T {
-        // Throw error if AssetStorage doesn't exist in the asset resources.
+    // Binding returned here to allow widgets to track loading.
+    pub fn get_asset<T>(&self, asset_handle: String) -> &Binding<Option<T>> {
+        // Create new asset if asset doesn't exist, and sets it to None.
+        ..
+    }
```
MrGVSV, line 38: This will also add the current widget to `binding.when_changed(...)`, right?

rfcs/rfc-2-font-measuring.md
```diff
@@ -42,13 +31,19 @@ pub struct KayakContext {
...
+    pub fn set_asset<T>(&mut self, asset_handle: String, asset: T) {
         ..
     }
 
-    // We want binding here because we need to allow widgets to track changes.
-    pub fn set_asset<T>(&mut self, asset: T) -> Binding<AssetHandle<T>> {
+    // Creates the AssetStorage<T> and stores it in assets
+    pub fn initialize_asset<T>(&mut self) {
         ..
     }
```
MrGVSV, line 48: Could we not handle this internally in `get_asset` or `set_asset` if the `AssetStorage` doesn't exist?

rfcs/rfc-2-font-measuring.md
```diff
@@ -0,0 +1,58 @@
...
+```rust
+pub struct KayakContext {
+    ..
+    // Stores AssetStorage generically.
+    assets: Resources,
+}
+
+impl KayakContext {
+    // Binding returned here to allow widgets to track loading.
+    pub fn get_asset<T>(&self, asset_handle: String) -> &Binding<Option<T>> {
```
MrGVSV, line 35: Actually, wouldn't we want something like:
```suggestion
    pub fn get_asset<T, I: Into<PathBuf>>(&self, asset_handle: I) -> &Binding<Option<T>> {
```

or:

```suggestion
    pub fn get_asset<T>(&self, asset_handle: impl Into<PathBuf>) -> &Binding<Option<T>> {
```

(and same for `set_asset`)

rfcs/rfc-2-font-measuring.md
StarArawn: Yes.

rfcs/rfc-2-font-measuring.md
StarArawn: Yeah I think the 2nd one makes sense.

rfcs/rfc-2-font-measuring.md
StarArawn: I was thinking this would mostly be for setting it up internally.

MrGVSV (approved):

PR #65, opened 2022-01-19
Added UI Detection

MrGVSV:
> ⚠️ This PR is based on #50 and requires that it be merged first.
>
> For reference, [here](https://github.com/StarArawn/kayak_ui/pull/65/files/ef0ff43a65b729eacb629a453b614445233778bd..5f7aaa7eefb10250112c309976a5452e2be64dbb) are the commits that are actually part of this PR.

## Goal

Add some way of letting the non-UI world know whether they should be registering clicks and other cursor events. For example, if the cursor is over a button, they probably don't also want to be interacting with the world beneath it.

The resultant API should look something like:

```rust
fn my_system(context: Res<BevyContext>, buttons: Res<Input<MouseButton>>, /* ... */) {
  if !buttons.just_pressed(MouseButton::Left) {
    // Mouse button wasn't pressed
    return;
  }

  if context.wants_cursor() {
    // Kayak is already interested in the mouse button
    return;
  }

  // Handle a click in the world
}
```

## Solution

Added some methods and logic to the `EventDispatcher` that will help inform users of the cursor's current status relative to the UI. The methods are also included in `KayakContext` and `BevyContext`.

> **Note:** This solution draws references from [egui](https://docs.rs/egui/0.16.1/src/egui/context.rs.html#726-761) for its API design

#### `contains_cursor()`

Simply returns true if the cursor is over a "valid" widget. Such widgets are defined by the fact that they have the ability to render their own content. Therefore, widgets specified with `RenderCommand::Empty`, `RenderCommand::Layout`, or `RenderCommand::Clip` will not affect the returned value.

#### `wants_cursor()`

If this returns true, it means that a widget either wants the cursor's input or is already using it. Right now, this includes all focusable widgets.

#### `has_cursor()`

Returns true when a widget is actively using the cursor's input. Right now, this applies for all widgets that affect `contains_cursor()` (but could maybe be limited to focusable widgets?). It maintains this state as long as the cursor's input remains pressed, even if it leaves the bounds of the widget (or the UI for that matter).

### Logic Placement

Right now the logic is found in `EventDispatcher::build_event_stream(...)` (the internal method for `EventDispatcher::process_events(...)`). This allows the most recent data to be available during the dispatch phase if a widget wants to access it from an event handler.

However, this prevents direct calls to `EventDispatcher::dispatch_events(...)` from updating the data. We could move the logic over to that method if we want to, but I'm not sure we need to. The `EventDispatcher::dispatch_events(...)` isn't currently called elsewhere, and if it were, we'd also need to update some other things that are updated in `EventDispatcher::build_event_stream(...)` anyways.

Just wanted to bring that part to attention in case we want to change how we handle it later (probably not in this PR).

### Widget Requirements

Another thing we could consider is how we determine a "valid" widget for all three methods. For `contains_cursor()`, should it allow some of the currently invalid widgets? Is there other criteria we should consider?

And for `has_cursor()`, should it only work on focusable widgets instead of just the "valid" widgets?

We could also see about allowing this logic to be controlled to some degree by a user. Though, that might be a bit more complex and should maybe go in a separate PR.

## Example

https://user-images.githubusercontent.com/49806985/150295478-a8ac9c56-67eb-49a0-8669-167b0ae8d728.mov

> The `world_interaction` example showcases how to properly handle world interactions in a way that considers Kayak. In the video, you can see that clicking over the UI does _not_ move the tile object.

### DIFF
--- a/bevy_kayak_ui/src/bevy_context.rs
+++ b/bevy_kayak_ui/src/bevy_context.rs
@@ -17,4 +17,41 @@ impl BevyContext {
 
         Self { kayak_context }
     }
+
+    /// Returns true if the cursor is currently over a valid widget
+    ///
+    /// For the purposes of this method, a valid widget is one which has the means to display a visual component on its own.
+    /// This means widgets specified with `RenderCommand::Empty`, `RenderCommand::Layout`, or `RenderCommand::Clip`
+    /// do not meet the requirements to "contain" the cursor.
+    pub fn contains_cursor(&self) -> bool {
+        if let Ok(kayak_context) = self.kayak_context.read() {
+            kayak_context.contains_cursor()
+        } else {
+            false
+        }
+    }
+
+    /// Returns true if the cursor may be needed by a widget or it's already in use by one
+    ///
+    /// This is useful for checking if certain events (such as a click) would "matter" to the UI at all. Example widgets
+    /// include buttons, sliders, and text boxes.
+    pub fn wants_cursor(&self) -> bool {
+        if let Ok(kayak_context) = self.kayak_context.read() {
+            kayak_context.wants_cursor()
+        } else {
+            false
+        }
+    }
+
+    /// Returns true if the cursor is currently in use by a widget
+    ///
+    /// This is most often useful for checking drag events as it will still return true even if the drag continues outside
+    /// the widget bounds (as long as it started within it).
+    pub fn has_cursor(&self) -> bool {
+        if let Ok(kayak_context) = self.kayak_context.read() {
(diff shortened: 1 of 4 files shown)


StarArawn (commented):
This looks great! Can we perhaps add a small demo? It would be neat if we could see it an action maybe with just a simple 3d cube scene?

MrGVSV:
> This looks great! Can we perhaps add a small demo? It would be neat if we could see it an action maybe with just a simple 3d cube scene?

Good idea! I'll add an example.

StarArawn (approved):

PR #66, opened 2022-01-22
Draggable Windows

MrGVSV:
## Objective

Add the ability to drag `Window` widgets around.

## Solution

Added the `draggable` prop to the `Window` widget, allowing it to be dragged from its title bar.

### Cursor Capturing

To add this functionality, I had to implement a way to capture all cursor events and pass them to the specified widget.

> See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture#example) for reference.

Therefore, I added the following methods (both on `EventDispatcher` and `KayakContext`):

* `capture_cursor(&mut self, index: Index)` - Captures all cursor events and sends them to the given widget instead
* `release_cursor(&mut self, index: Index) -> bool` - Releases the capture if the index matches
* `force_release_cursor(&mut self) -> Option<Index>` - Forcibly releases the capture, regardless of who owns it

## Additional Changes

Also added `CursorEvent` to all cursor-related variants in `EventType`. This was added to make accessing cursor related data right from the event.

```rust
pub struct CursorEvent {
  pub pressed: bool,
  pub just_pressed: bool,
  pub just_released: bool,
  pub position: (f32, f32),
}
```

This means all cursor event handlers needed to change from:

```rust
match event.event_type {
  EventType::Click => {/* ... */}
  EventType::MouseIn => {/* ... */}
  EventType::Hover => {/* ... */}
  _ => {}
}
```

to:

```rust
match event.event_type {
  EventType::Click(_) => {/* ... */}
  EventType::MouseIn(..) => {/* ... */}
  EventType::Hover(data) => {/* ... */}
  _ => {}
}
```

## Example

Added the `draggable` prop to the `counter` and `world_interaction` examples.

https://user-images.githubusercontent.com/49806985/150621137-d9dc3a11-47d6-4814-bc59-83700c44a977.mov

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Widget's can create their own state and will re-render when that state changes.
 fn Counter(context: &mut KayakContext) {
     let (count, set_count, ..) = use_state!(0i32);
     let on_event = OnEvent::new(move |_, event| match event.event_type {
-        EventType::Click => set_count(count + 1),
+        EventType::Click(..) => set_count(count + 1),
         _ => {}
     });
 
--- a/examples/counter.rs
+++ b/examples/counter.rs
@@ -31,7 +31,7 @@ fn Counter(context: &mut KayakContext) {
 
     let (count, set_count, ..) = use_state!(0i32);
     let on_event = OnEvent::new(move |_, event| match event.event_type {
-        EventType::Click => set_count(count + 1),
+        EventType::Click(..) =>  set_count(count + 1),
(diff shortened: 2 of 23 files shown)


MrGVSV:
Just changed the `Window` size for the `world_interaction` example. I think being able to move it around should work pretty well for that example (instead of the full width `Window`).

StarArawn (approved):

PR #67, opened 2022-01-22
Border

StarArawn:


bevy_kayak_ui/src/render/unified/shader.wgsl
```diff
@@ -51,39 +51,26 @@ var image_sampler: sampler;
...
-    var top_left: vec2<f32> = position + radius;
-    var bottom_right: vec2<f32> = top_left + inner_size;
-
-    var top_left_distance: vec2<f32> = top_left - frag_coord;
-    var bottom_right_distance: vec2<f32> = frag_coord - bottom_right;
-
-    var dist = max(max(top_left_distance, bottom_right_distance), vec2<f32>(0.0));
-
-    return length(dist);
+fn sdRoundBox(p: vec2<f32>, b: vec2<f32>, r: f32) -> f32
```
MrGVSV, line 54: Maybe add a comment clarifying what `p`, `b`, and `r` are? I'm not super familiar with shaders so something like that would be helpful if I (or anyone else) need to use this function.

MrGVSV (approved):
Looks good to me! Only comment is regarding documentation but isn't blocking.

bevy_kayak_ui/src/render/unified/shader.wgsl
StarArawn: P is the position for the given fragment. I use local position since that seems to be the easiest to calculate. B is the size of the rounded box and R is the radius of the circle at the corner.

MrGVSV: I meant in the code 😅

StarArawn: Done!

PR #68, opened 2022-02-01
Swapped us over to a `KayakContextRef` which better handles some things.

StarArawn:


src/widgets/inspector.rs
```diff
@@ -35,41 +35,41 @@ pub fn Inspector() {
     let (id, _) = last_clicked_value.into_raw_parts();
 
     let mut parent_id_move = None;
-    if let Some(layout) = context.widget_manager.get_layout(&last_clicked_value) {
-        if let Some(node) = context.widget_manager.get_node(&last_clicked_value) {
-            let mut data = Vec::new();
+    // if let Some(layout) = context.widget_manager.get_layout(&last_clicked_value) {
```
StarArawn, line 38: I'll fix this. I just have to expose a bit more stuff.

kayak_core/src/fragment.rs
```diff
@@ -39,23 +39,15 @@ impl Widget for Fragment {
...
-    fn render(&mut self, context: &mut KayakContext) {
+    fn render(&mut self, context: &mut KayakContextRef) {
         let parent_id = self.get_id();
-        let tree = crate::WidgetTree::new();
-        tree.add(parent_id, None);
-
         if let Some(children) = self.children.take() {
-            children(tree.clone(), Some(parent_id), context);
+            let mut context = KayakContextRef::new(&mut context.context, Some(parent_id));
+            children(Some(parent_id), &mut context);
```
MrGVSV, line 46: Would we want to consider adding another method to `KayakContextRef` to do this for a user? It might be a common enough pattern.

MrGVSV: Or is this logic exclusive to `Fragment`?

kayak_core/src/fragment.rs
```diff
@@ -39,23 +39,15 @@ impl Widget for Fragment {
...
         }
 
-        // Consume the widget tree taking the inner value
-        let tree = tree.take();
-
-        // Evaluate changes to the tree.
-        let changes = context.widget_manager.tree.diff_children(&tree, parent_id);
-
-        context.widget_manager.tree.merge(&tree, parent_id, changes);
+        // Note: No need to do anything here with this KayakContextRef.
```
MrGVSV, line 51: Is there a reason we don't call `context.commit()` here? If there is, it might be helpful to leave a comment stating why.

kayak_render_macros/src/arc_function.rs
```diff
@@ -4,34 +4,11 @@ use quote::quote;
...
-        context
-            .widget_manager
-            .create_widget(#index, #widget_name, #parent);
-        #tree_add
-        if should_rerender {
-            let mut child_widget = context.widget_manager.take(child_id);
-            child_widget.render(context);
-            context.widget_manager.repossess(child_widget);
-        }
+        context.add_widget(#widget_name, #index);
```
MrGVSV, line 12: Amazing 😯

MrGVSV (approved):
It's looking really good!

kayak_core/src/fragment.rs
StarArawn: Because fragments are a bit weird the logic here is that we pass in a new context ref into the closure. I'm not sure we need to provide a different API. Fragments currently are the only way to resolve children I think..

kayak_core/src/fragment.rs
StarArawn: Fragments are static so nothing really changes here. The fragment doesn't actually own the children rather the widget above the fragment does.

kayak_core/src/fragment.rs
MrGVSV: Makes sense, thanks!

MrGVSV (approved):

PR #69, opened 2022-02-01
Added `WidgetProps`

MrGVSV:
Added `WidgetProps` and its derive macro, according to RFC 1: #56.

Tasks:
- [x] Add `WidgetProps`
- [x] Add derive macro
- [x] Integrate with `Widget` (likely requires updating the rest of the codebase to account for it)


kayak_core/src/vec.rs
```diff
@@ -1,71 +1,103 @@
...
 
-use crate::{context::KayakContext, context_ref::KayakContextRef, styles::Style, Index, Widget};
+use crate::{context_ref::KayakContextRef, styles::Style, Index, Widget, Children, OnEvent, WidgetProps};
 
 #[derive(Derivative)]
-#[derivative(Debug, PartialEq, Clone, Default)]
-pub struct VecTracker<T> {
-    pub id: Index,
+#[derivative(Default, Debug, PartialEq, Clone)]
+pub struct VecTrackerProps<T> {
```
StarArawn, line 7: We should use a derive here that adds the other fields. :)

kayak_core/src/widget.rs
```diff
@@ -1,26 +1,113 @@
...
     fn render(&mut self, context: &mut KayakContextRef);
+    fn get_name(&self) -> &'static str;
+    fn on_event(&mut self, context: &mut KayakContext, event: &mut Event);
 }
 
-impl as_any::Downcast for dyn Widget {}
-impl as_any::Downcast for dyn Widget + Send {}
-impl as_any::Downcast for dyn Widget + Sync {}
-impl as_any::Downcast for dyn Widget + Send + Sync {}
+pub trait Widget: std::fmt::Debug + Clone + Default + PartialEq + AsAny + Send + Sync {
```
StarArawn, line 27: Wait how can we do `+ Clone` here? Pretty sure that wasn't allowed. 🤔

kayak_core/src/widget_manager.rs
```diff
@@ -67,7 +57,7 @@ impl WidgetManager {
         }
     }
 
-    pub fn create_widget<T: Widget + PartialEq + Default + Clone + 'static>(
+    pub fn create_widget<T: Widget + 'static>(
```
StarArawn, line 60: Much cleaner! 👍

kayak_core/src/widget.rs
StarArawn, line 27: So is `Widget` here not object safe, but it doesn't matter because it implements `BaseWidget` which is? 🤔

MrGVSV: It isn't allowed if we want `Widget` to be object-safe. Using `BaseWidget` allows us to side-step that requirement by turning `Self::Props` into `dyn WidgetProps`.

In theory, this allows implementors of widget to work with their concrete types, but allows our internals to just work with the traits.

MrGVSV: Yeah that was my idea. I figure it's a simple way of allowing users to work with their concrete types while allowing us to use this trait as a dynamic trait object.

StarArawn (changes requested):
Looks pretty good just a couple small things!

src/widgets/background.rs
```diff
@@ -1,16 +1,32 @@
...
     styles::{Style, StyleProp},
     widget, Children, Fragment,
 };
 
+#[derive(WidgetProps, Default, Debug, PartialEq, Clone)]
+pub struct BackgroundProps {
+    #[prop_field(Styles)]
+    pub styles: Option<Style>,
+    #[prop_field(Children)]
+
```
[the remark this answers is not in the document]
MrGVSV: Yeah haha that was the result of a mass-find-and-replace. I'll clean it all up when this PR is just about finished.

src/widgets/background.rs
```diff
@@ -1,16 +1,32 @@
...
 #[widget]
-pub fn Background(children: Children, styles: Option<Style>) {
-    if styles.is_none() {
-        *styles = Some(Style::default())
+pub fn Background(props: BackgroundProps) {
+    if props.styles.is_none() {
+        props.styles = Some(Style::default())
     }
-    styles.as_mut().unwrap().render_command = StyleProp::Value(RenderCommand::Quad);
+    props.styles.as_mut().unwrap().render_command = StyleProp::Value(RenderCommand::Quad);
```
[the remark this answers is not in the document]
MrGVSV: And when we add the widget contexts, it will be like a regular function haha (no more hidden `self` and `context`).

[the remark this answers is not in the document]
StarArawn: That's what I love!! This turns the api from mess to nice. 😁

kayak_render_macros/src/function_component.rs
```diff
@@ -13,203 +15,231 @@ impl Default for WidgetArguments {
...
-    TokenStream::from(quote! {
-        use #kayak_core::derivative::*;
+    let block = f.block;
+    let vis = f.vis;
 
-        #[derive(Derivative)]
-        #[derivative(Default, Debug, PartialEq, Clone)]
+    let kayak_core = get_core_crate();
+
+    // TODO: See if this is still needed. If not, remove it
```
StarArawn, line 66: I think it's safe to remove this now!!

StarArawn (changes requested):

StarArawn (approved):

PR #70, opened 2022-02-02
Fixed crashes when no context existed.

StarArawn:


MrGVSV (approved):

examples/bevy_state.rs
```diff
@@ -0,0 +1,100 @@
...
+    MainMenu,
+    Options,
+    Play,
+}
+
+fn create_main_menu(mut commands: Commands) {
+    let context = BevyContext::new(|context| {
+        render! {
+            <App>
+                <Text content={"Main Menu".to_string()} size={32.0} />
```
MrGVSV, line 22: Might be worth putting instructions to press space to switch to the next state. Just to flesh it out a bit more.

MrGVSV (approved):
Looks good! Minor suggestion but not super necessary to merge

bevy_kayak_ui/src/lib.rs
```diff
@@ -36,69 +36,86 @@ pub(crate) fn to_bevy_color(color: &kayak_core::color::Color) -> Color {
...
-    bevy_context: Res<BevyContext>,
-    windows: Res<Windows>,
-    mut cursor_moved_events: EventReader<CursorMoved>,
-    mut mouse_button_input_events: EventReader<MouseButtonInput>,
-    mut char_input_events: EventReader<ReceivedCharacter>,
-    mut keyboard_input_events: EventReader<KeyboardInput>,
-) {
-    let window_size = if let Some(window) = windows.get_primary() {
-        Vec2::new(window.width(), window.height())
+pub fn process_events(world: &mut World) {
```
StarArawn, line 49: @MrGVSV 
Maybe we should just move process_events into update?

bevy_kayak_ui/src/render/unified/font/font_mapping.rs
```diff
@@ -32,6 +32,11 @@ impl FontMapping {
         }
     }
 
+    pub(crate) fn mark_all_as_new(&mut self) {
```
StarArawn, line 35: This just lets us reprocess fonts into the kayak context.

examples/bevy_state.rs
```diff
@@ -0,0 +1,120 @@
...
+                <Text content={"Options".to_string()} size={32.0} />
+                <StateSwitcher />
+            </App>
+        }
+    });
+
+    commands.insert_resource(context);
+}
+
+fn create_play_menu(mut commands: Commands) {
```
StarArawn, line 75: The more I think about this the more I dislike it. I think users should instead use scenes here. We should think about how we want to handle that.

src/widgets/text.rs
```diff
@@ -48,30 +50,36 @@ pub fn Text(props: TextProps) {
...
         };
 
-    let render_command = RenderCommand::Text {
-        content: content.clone(),
-        size,
-        parent_size,
-        line_height: line_height.unwrap_or(size * 1.2),
-        font: font_name.clone().unwrap_or("Roboto".into()),
-    };
+    if should_render {
```
StarArawn, line 67: This prevents rendering text until all of the conditions are met. I found this useful incases where font has loaded and we don't have the parent layout. It keeps the text from jumping around. :)

bevy_kayak_ui/src/lib.rs
MrGVSV: Seems reasonable.

examples/bevy_state.rs
MrGVSV: I think it's okay to do for now. We could maybe explore scenes in another PR/RFC?

kayak_core/src/context_ref.rs
```diff
@@ -260,6 +260,14 @@ impl<'a> KayakContextRef<'a> {
             .tree
             .merge(&tree, self.current_id.unwrap_or_default(), changes);
     }
+
+    /// This function is specifically for text rendering which needs to re-render when
+    /// it's parent layout is calculated.
+    pub fn mark_dirty(&mut self, widget_id: &Index) {
```
MrGVSV, line 266: Why can't we just get `widget_id` from `current_id`?

StarArawn: Because I didn't think of it 👍

PR #74, opened 2022-02-07
Improved Styles

MrGVSV:
## Objective

This PR tackles two main issues:

1. `StyleProp::Inherit` was not working and attempting to resolve the style resulted in a panic
2. There was no easy way of merging styles

### Inheritance

Solving inheritance was rather straightforward. We basically just needed a way to process nodes in the order they are added to the tree. Since we were using a `HashSet`, order was not guaranteed. The fix was to add the [`indexmap`](https://crates.io/crates/indexmap) crate as a dependency and use its drop-in replacement, `IndexSet`.

That pretty much did it surprisingly. The last change was to try and fetch the parent's resolved style before fetching their prop style (or default if not available).

I am not sure how robust of a solution this is, but initial tests show that it works (the Text in the Textbox widget, for example, can now inherit its color from the TextBox).

### Style Helpers

Managing styles can be difficult, especially if you're trying to pass and process many of them. This PR also addresses this by adding some helper methods.

First, though, it adds a new variant: `StyleProp::Unset`. This variant just means that a prop has not yet been set. This is distinct in `StyleProp::Default` in that the intention is to allow others to set its value (in some cases, we may actually want `StyleProp::Default`).

> This variant is also the new default value used in `StyleProp::default()` and `Style::default()`.

With this new variant, we can easily apply styles additively.

For example, instead of doing something like:

```rust
let styles = Style {
  // Required
  render_command: StyleProp::Value(RenderCommand::Layout),

  // Optional with defaults
  color: if matches!(props.styles.color, StyleProp::Unset) {
      props.styles.color.clone()
    } else {
      StyleProp::Value(Color::White)
    },
  width: if matches!(props.styles.width, StyleProp::Unset) {
      props.styles.width.clone()
    } else if matches!(other_styles.width, StyleProp::Unset) {
      other_styles.width.clone()
    } else {
      StyleProp::Value(Units::Pixels(25.0))
    },
  height: if matches!(props.styles.height, StyleProp::Unset) {
      props.styles.height.clone()
    } else {
      StyleProp::Value(Units::Pixels(50.0))
    },
  ..Default::default()
};
```

We can do this:

```rust
let styles = Style::default()
  // Our required styles
  .with_style(Style {
    render_command: StyleProp::Value(RenderCommand::Layout),
    ..Default::default()
  })

  // Styles from props
  .with_style(&props.styles)
  // Maybe some other styles
  .with_style(&other_styles)

  // Our defaults
  .with_style(Style {
    color: StyleProp::Value(Color::White),
    width: StyleProp::Value(Units::Pixels(25.0)),
    height: StyleProp::Value(Units::Pixels(50.0)),
    ..Default::default()
  });
```

> `with_style` can take any of the following: `Style`, `&Style`, `Option<Style>`, and `&Option<Style>`, making it very versatile.

Which is just a bit more readable (or at least more generic/concise).

There's also the `apply` method which does the same thing, but to a mutable self (no return).

#### Into StyleProp

This PR also adds a blanket `From` impl for `StyleProp`. This is just a quality of life thing to save us some keystrokes and imports:

```rust
let styles = Style::default()
  // Our required styles
  .with_style(Style {
    render_command: RenderCommand::Layout.into(),
    ..Default::default()
  })

  // Styles from props
  .with_style(&props.styles)
  // Maybe some other styles
  .with_style(&other_styles)

  // Our defaults
  .with_style(Style {
    color: Color::White.into(),
    width: Units::Pixels(25.0).into(),
    height: Units::Pixels(50.0).into(),
    ..Default::default()
  });
```

#### Select StyleProp

For smaller merges, we might not need to merge the entire `Style` object. Instead, we can do something similar on a smaller scale with `StyleProp::select(...)`:

```rust
let styles = Style {
  // Required
  render_command: StyleProp::Value(RenderCommand::Layout),
  // Optional with defaults
  color: StyleProp::select(&[
    &props.styles.color, 
    &other_styles.color, 
    &StyleProp::Value(Color::White)
  ]).clone(),
  ..Default::default()
};
```

#### `define_styles!`

The last change is an internal one. Basically, I made a macro (`define_styles!`) that can be used to generate the `Style` struct along with other implementation methods. The main purpose of this was just to reduce the number of errors caused by adding a new field but not updating all the methods to account for the change.

## Feedback

Any part of this PR can be renamed, removed, changed, etc. Let me know if there are any thoughts!

### DIFF
--- a/kayak_core/src/styles.rs
+++ b/kayak_core/src/styles.rs
@@ -8,7 +8,7 @@ pub enum StyleProp<T: Default + Clone> {
     /// This prop is unset, meaning its actual value is not determined until style resolution,
     /// wherein it will be set to the property's default value.
     ///
-    /// When [merging](Style::merge) styles, only style properties of this type may be
+    /// When [merging](Style::apply) styles, only style properties of this type may be
     /// overwritten.
     Unset,
     /// Like [StyleProp::Unset], properties of this type wait until style resolution for their
@@ -78,7 +78,10 @@ macro_rules! define_styles {
         impl $name {
 
             /// Returns a `Style` object where all fields are set to [`StyleProp::Default`]
-            pub(crate) fn defaulted() -> Self {
+            ///
+            /// This should only be used when default properties are required or desired. Otherwise, you
+            /// may be better off using `Style::default` (which sets all properties to [`StyleProp::Unset`]).
+            pub(crate) fn new_default() -> Self {
                 Self {
                     $($field: StyleProp::Default),*
                 }
@@ -93,15 +96,25 @@ macro_rules! define_styles {
                  )*
             }
 
-            /// Merges two `Style` objects
+            /// Applies a `Style` over this one
             ///
-            /// Values from `other` are applied to any field that is marked as [`StyleProp::Unset`]
-            pub fn merge(&mut self, other: &Self) {
-                 $(
-                     if matches!(self.$field, StyleProp::Unset) {
-                         self.$field = other.$field.clone();
-                     }
-                 )*
+            /// Values from `other` are applied to any field in this one that is marked as [`StyleProp::Unset`]
+            pub fn apply<T: AsRefOption<Style>>(&mut self, other: T) {
+                 if let Some(other) = other.as_ref_option() {
+                     $(
+                         if matches!(self.$field, StyleProp::Unset) {
+                             self.$field = other.$field.clone();
+                         }
+                     )*
+                 }
+            }
+
+            /// Applies the given style and returns the updated style
+            ///
+            /// This is simply a builder-like wrapper around the [`Style::apply`] method.
+            pub fn with_style<T: AsRefOption<Style>>(mut self, other: T) -> Self {
+                self.apply(other);
+                self
             }
         }
     };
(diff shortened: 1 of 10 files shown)


kayak_core/src/styles.rs
```diff
@@ -5,8 +5,18 @@ use crate::{color::Color, render_command::RenderCommand};
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum StyleProp<T: Default + Clone> {
+    /// This prop is unset, meaning its actual value is not determined until style resolution,
+    /// wherein it will be set to the property's default value.
+    ///
+    /// When [applying](Style::apply) styles, only style properties of this type may be
+    /// overwritten.
+    Unset,
```
MrGVSV, line 13: I also considered `Initial`, but `Unset` just seems more clear to me.

kayak_core/src/styles.rs
```diff
@@ -15,7 +25,7 @@ where
     T: Default + Clone,
 {
     fn default() -> Self {
-        Self::Default
+        Self::Unset
     }
```
MrGVSV, line 29: If we don't want `Style::default` to return `StyleProp::Unset` fields, we could replace this with a `Style::unset()` method. But I prefer implementing it on `Default` since it allows us to call `unwrap_or_default()` with the unset properties.

StarArawn (approved):
I'm a little worried about the styles macro but it makes sense.

MrGVSV:
> I'm a little worried about the styles macro but it makes sense.

We can get rid of it if you'd prefer, I added it mainly to reduce the amount of code we need to write/maintain and to reduce potential bugs.

PR #75, opened 2022-02-07
Added `cursor` style

### LINKED ISSUE #6: Change Cursor
Right now bevy doesn't expose winit's `set_cursor_icon` function. We need to enable that functionality in bevy to show different cursors. Then we can hook into the events to change the cursor icon.

MrGVSV:
## Objective

Add a simple method for setting the cursor's icon.

## Solution

Added a `CursorIcon` enum that maps one-to-one to Bevy's (which itself comes from winit). Then added a new style prop called `cursor`, which initializes to `StyleProp::Inherit`.

### Examples

Added some cursor icons to certain elements:

* Text: `CursorIcon::Text`
* TextBox: `CursorIcon::Text`
* Button: `CursorIcon::Hand`
* Window (Title Bar): `CursorIcon::Grab` and `CursorIcon::Grabbing`

> These can be removed if they're too much. They were mainly for testing/demonstration anyways.

### Screenshots

https://user-images.githubusercontent.com/49806985/152857083-029433c9-5601-4483-8568-7bc95c0be73d.mov

## Context

Closes #6.


bevy_kayak_ui/src/lib.rs
```diff
@@ -40,8 +42,15 @@ pub fn update(world: &mut World) {
...
             context.set_global_state(std::mem::take(world));
             context.render();
-            *world = context.take_global_state::<World>().unwrap()
+            *world = context.take_global_state::<World>().unwrap();
+
+            if let Some(ref mut windows) = world.get_resource_mut::<Windows>() {
+                if let Some(window) = windows.get_primary_mut() {
+                    window.set_cursor_icon(convert_cursor_icon(context.cursor_icon()));
+                }
+            }
```
MrGVSV, line 51: Alternatively, we could maybe register a callback for when the cursor icon changes. But this was the simplest method.

kayak_core/src/context.rs
```diff
@@ -557,4 +560,27 @@ impl KayakContext {
...
+        }
+
+        let hovered = self.event_dispatcher.hovered.unwrap();
+        if let Some(node) = self.widget_manager.nodes.get(hovered) {
+            if let Some(node) = node {
+                let icon = node.resolved_styles.cursor.resolve();
+                self.cursor_icon = icon;
+            }
+        }
+    }
```
MrGVSV, line 585: Should this be moved outside of `KayakContext`? Does it make more sense in `EventDispatcher`? Or a new structure entirely?

kayak_core/src/context.rs
```diff
@@ -27,6 +27,7 @@ pub struct KayakContext {
     widget_state_lifetimes:
         HashMap<crate::Index, HashMap<crate::flo_binding::Uuid, Box<dyn crate::Releasable>>>,
     widget_states: HashMap<crate::Index, resources::Resources>,
+    cursor_icon: CursorIcon,
```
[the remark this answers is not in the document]
StarArawn: I think it's fine here. 🙂

kayak_core/src/context.rs
StarArawn: Let's leave it for now. I think maybe we can separate out some logic with a few traits maybe.

StarArawn (approved):

PR #76, opened 2022-02-08
Empty props

MrGVSV:
## Objective

Make widgets that don't require props, easier to write. This comes in handy for fully contained widgets or widgets written for examples.

## Solution

Implemented `WidgetProps` over the unit type (`()`). Also made it so that functional widgets defined with zero arguments use this empty prop type.


kayak_core/src/widget.rs
```diff
@@ -120,3 +120,23 @@ where
 }
 
 impl<T> SealedWidget for T where T: Widget {}
+
+impl WidgetProps for () {
```
MrGVSV, line 124: We could also replace this with a concrete `EmptyProps` type if preferred.

StarArawn: This is great! 🙂

StarArawn (approved):

PR #80, opened 2022-02-11
Add General Documentation

MrGVSV:
## Objective

Add documentation to most outwards-facing fields and methods. This documentation is not meant to cover ever detail (we need to leave some stuff for the book...). The point of this is to make sure users who prefer using docs.rs can do so easily, as well as users who prefer reading documentation directly in their editor.

Here is a (mildly long) list of things that currently need better documentation:

- [x] Context
  - [x] `KayakContextRef`
  - [x] `KayakContext` (most of this will be copied from `KayakContextRef` once that's finalized)
- [x] Bevy
  - [x] `BevyContext` (on the struct itself)
  - [x] `FontMapping`
  - [x] `ImageManager`
- [x] Widgets
  - [x] `Style`
  - [x] `Node`
  - [x] `WidgetProps`
- [x] Events
  - [x] `InputEvent`
  - [x] `EventType`
  - [x] `Handler`
- [x] Macros
  - [x] `constructor!`
  - [x] `render!`
  - [x] `#[widget]`
  - [x] `#[derive(WidgetProps)]`
- [x] Built-in Widgets (really just explain what they do and when their needed)
  - [x] `App`
    - [x] `AppProps`
  - [x] `Background`
    - [x] `BackgroundProps`
  - [x] `Button`
    - [x] `ButtonProps`
  - [x] `ChangeEvent` (can this be replaced by a `Handler`)
  - [x] `Clip`
    - [x] `ClipProps`
  - [x] `Element`
    - [x] `ElementProps`
  - [x] ~`Focus`~ Removed since it's only used internally
  - [x] `Fold`
    - [x] `FoldProps`
  - [x] `Fragment`
    - [x] `FragmentProps`
  - [x] `If`
    - [x] `IfProps`
  - [x] `Image`
    - [x] `ImageProps`
  - [x] `Inspector`
    - [x] `InspectorProps`
  - [x] `NinePatch`
    - [x] `NinePatchProps`
  - [x] `OnChange`
  - [x] `Text`
    - [x] `TextProps`
  - [x] `TextBox`
    - [x] `TextBoxProps`
  - [x] `TooltipConsumer`
    - [x] `TooltipConsumerProps`
  - [x] `TooltipData`
  - [x] `TooltipProvider`
    - [x] `TooltipProviderProps`
  - [x] `VecTracker`
    - [x] `VecTrackerProps`
  - [x] `Window`
    - [x] `WindowProps`

> `WidgetManager` has been skipped as it shouldn't really be a public facing API and will likely be slowly privatized.

## Feedback & Contributing
Let me know if anything is missing, something _shouldn't_ be included, or anything else like that! And if other contributors want to help, even if only for a single item, please feel free to make a PR to this branch!

For now, this PR will be left in draft form until it's near-ready to be merged, however, reviews are still very much welcome.


kayak_core/src/context.rs
```diff
@@ -364,20 +468,79 @@ impl KayakContext {
...
+    ///   let bound_count = bind(MyCount(123));
+    ///   context.bind(&bound_count);
+    ///   context.set_global(bound_count);
+    /// }
+    /// ```
+    pub fn set_global<T: resources::Resource>(&mut self, value: T) {
+        self.global_state.insert(value);
     }
 
-    pub fn get_global_state<T: resources::Resource>(
```
MrGVSV, line 371: Thoughts on renaming this? Calling it `global_state` might be a bit confusing since it's only a "state" (as we define it elsewhere) when the given resource is a `Binding<T>` and a widget actually binds to it.

kayak_core/src/fragment.rs
```diff
@@ -2,12 +2,26 @@ use crate::{
...
+///
+/// __Type:__ [`FragmentProps`]
+///
+/// | Common Prop | Accepted |
+/// | :---------: | :------: |
+/// | `children`  | ✅        |
+/// | `styles`    | ✅        |
+/// | `on_event`  | ❌        |
+/// | `focusable` | ❌        |
+///
```
MrGVSV, line 24: Thoughts on this layout? For built-in widgets, I think it's nice to show right on the widget what common props it accepts. Although, I'm open to alternatives (as well as just not showing that information).

kayak_core/src/styles/mod.rs
```diff
@@ -141,23 +155,91 @@ macro_rules! define_styles {
...
+    /// };
+    ///
+    /// let style = Style::default() // <- Initializes all fields as `StyleProp::Unset`
+    ///   // Applied first (sets any `StyleProp::Unset` fields)
+    ///   .with_style(&style_a)
+    ///   // Applied second (sets any remaining `StyleProp::Unset` fields)
+    ///   .with_style(&style_b);
+    /// ```
     #[derive(Debug, Default, Clone, PartialEq)]
     pub struct Style {
```
MrGVSV, line 195: Should we also add a [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html) attribute? I imagine this struct could change a fair amount more (although I hope no one is manually writing out all fields...).

kayak_core/src/context.rs
StarArawn: Yeah I think this sounds fine. :)

kayak_core/src/context.rs
```diff
@@ -457,12 +642,33 @@ impl KayakContext {
...
+    /// * `f`: The function to call with the given system parameter
+    ///
+    /// # Examples
+    ///
+    /// ```ignore
+    /// use bevy::prelude::Res;
+    ///
+    /// struct MyCount(i32);
+    ///
+    ///#[widget]
```
StarArawn, line 661: Can we show a second example here with multiple parameters?

kayak_core/src/context_ref.rs
```diff
@@ -138,13 +333,26 @@ impl<'a> KayakContextRef<'a> {
         self.context.query_world::<T, F, R>(f)
     }
 
+    /// Get a stored asset with the given asset key
+    ///
+    /// # Arguments
+    ///
+    /// * `key`: The asset key
```
StarArawn, line 340: Probably a good idea to discuss T here?

kayak_core/src/fragment.rs
```diff
@@ -2,12 +2,26 @@ use crate::{
...
 
+/// The base widget, used to actually build and render children
+///
+/// # Props
+///
+/// __Type:__ [`FragmentProps`]
+///
+/// | Common Prop | Accepted |
+/// | :---------: | :------: |
+/// | `children`  | ✅        |
```
StarArawn, line 20: pretty. 🤣

kayak_core/src/styles/mod.rs
```diff
@@ -141,23 +155,91 @@ macro_rules! define_styles {
...
+        ///
+        /// Currently, this controls all border sides.
+        ///
+        /// Only applies to widgets marked [`RenderCommand::Quad`]
         pub border_color: StyleProp<Color>,
+        /// The radius of the corners (in pixels)
+        ///
+        /// The order is (Top, Right, Bottom, Left).
+        ///
+        /// Only applies to widgets marked [`RenderCommand::Quad`] and [`RenderCommand::Image`]
```
StarArawn, line 210: Ideally we should allow both quads and images to have borders and background colors, but this is accurate right now. 👍

StarArawn (changes requested):

kayak_core/src/context.rs
MrGVSV: How's that?

kayak_core/src/context_ref.rs
MrGVSV: Discussed!

kayak_core/src/styles/mod.rs
StarArawn: Lets do that. :)

kayak_core/src/context.rs
StarArawn: Looks good!

kayak_core/src/styles/mod.rs
MrGVSV: Nevermind, just tried it and apparently it doesn't support the update syntax. PReventing us from doing:

```rust
let styles = Style {
  color: Color::BLACK.into(),
  ..Default::default()
}
```

And instead forcing us to do:

```rust
let mut styles = Style::default();
styles.color = Color::BLACK.into();
```

Which makes the new builder patterns a little more awkward to use:

```rust
let mut user_styles = Style::default();
user_styles.color = Color::BLACK.into();
// ...
```

StarArawn: Just keep it as is! 👍

StarArawn (approved):

PR #81, opened 2022-02-12
Added QoL improvements to styling

MrGVSV:
## Objective

Make configuring style a little more intuitive and a little less verbose.

## Solution

Added the `Edge<T>` struct which replaces the `(f32, f32, f32, f32)`. This should help reduce the confusion of which tuple field means what, as well as make defining them much more compact due to convenience methods on `Edge<T>`.

Examples of this in action:

```rust
// Old
let border = (10.0, 10.0, 10.0, 10.0);
// New
let border = Edge::all(10.0);

// Old
let border = (10.0, 5.0, 10.0, 5.0);
// New
let border = Edge::all(10.0, 5.0);

// Old
let bottom = border.3;
// New
let bottom = border.bottom;

// Old
border = (
  border.0 * dpi,
  border.1 * dpi,
  border.2 * dpi,
  border.3 * dpi,
);
// New
border *= dpi;
```

Furthermore, `Edge<T>` is generic, allowing it to be used for things like border color when we add the ability to set border color individually.

Also add a similar struct for corners (aptly named `Corner<T>`).

### Changed Style Props

* `border` (now uses `Edge<f32>`)
* `border_radius` (now uses `Corner<f32>`)

### Added Style Props

* `padding` - Controls padding
* `offset` - Controls margin/position
* ~`col_gap`~ `col_between` - Adds space between children horizontally
* ~`row_gap`~ `row_between` - Adds space between children vertically

### Additional Changes

Also restructured styles stuff into its own directory to help keep things clean.


kayak_core/src/styles/mod.rs
```diff
@@ -138,22 +143,29 @@ define_styles! {
...
         pub background_color : StyleProp<Color>,
         pub border_color: StyleProp<Color>,
-        pub border_radius: StyleProp<(f32, f32, f32, f32)>,
-        pub border: StyleProp<(f32, f32, f32, f32)>,
+        pub border_radius: StyleProp<Edge<f32>>,
+        pub border: StyleProp<Edge<f32>>,
         pub bottom: StyleProp<Units>,
         pub color: StyleProp<Color>,
+        /// The spacing between child widgets along the horizontal axis
+        pub col_gap: StyleProp<Units>,
```
MrGVSV, line 151: Should `col_gap` and `row_gap` be renamed to match Morphorm's naming (`col_between` and `row_between`)? I chose "gap" to be more consistent with CSS, but I'm okay with either honestly.

kayak_core/src/styles/mod.rs
```diff
@@ -139,21 +144,28 @@ define_styles! {
         pub background_color : StyleProp<Color>,
         pub border_color: StyleProp<Color>,
         pub border_radius: StyleProp<(f32, f32, f32, f32)>,
```
MrGVSV, line 146: Should `border_radius` have a similar struct to `Edge` (probably not generic) to provide convenience methods and be more readable (i.e. with obvious top-left, top-right, bottom-left, bottom-right fields)? Maybe `Corner`?

kayak_core/src/styles/mod.rs
```diff
@@ -139,21 +144,28 @@ define_styles! {
...
         pub min_width: StyleProp<Units>,
+        /// The inner padding between the edges of this widget and its children
+        ///
+        /// This property has lower precedence than its more specific counterparts
+        /// (`padding_top`, `padding_right`, `padding_bottom`, and `padding_left`), allowing it
+        /// to be overridden.
+        ///
+        /// A child with their own padding properties set to anything other than `Units::Auto` will
+        /// override the padding set by this widget.
+        pub padding: StyleProp<Edge<Units>>,
```
MrGVSV, line 168: If we want to keep `padding`, we probably want to add one that covers `top`, `right`, `bottom`, and `left` as well, right? If so, what should that be named? `offset`? `position`?

kayak_core/src/styles/mod.rs
StarArawn: I think we should stay consistent with morphorm, I'm not sure I've always done that but it might help with documentation and type sharing in the future. 🤔

kayak_core/src/styles/mod.rs
StarArawn: Not sure I follow? Isn't padding top, right, bottom and left already? Or do you mean separate properties?

MrGVSV: > Not sure I follow? Isn't padding top, right, bottom and left already? Or do you mean separate properties?

Sorry haha yeah that wasn't very clear.

Right now `padding` covers `padding_top`, `padding_right`, `padding_bottom`, and `padding_left`. My question is, do we want something similar to cover `top`, `right`, `bottom`, and `left`?

I ask mainly because those latter properties act as both positioning and "margin" within morphorm it seems. And if we do include a property for them, should it be named something like `offset` or `position` or something else?

kayak_core/src/styles/mod.rs
```diff
@@ -138,22 +145,40 @@ define_styles! {
...
+        /// The positional offset from this widget's default position
+        ///
+        /// This property has lower precedence than its more specific counterparts
+        /// ([`top`](Self::top), [`right`](Self::right), [`bottom`](Self::bottom), and [`left`](Self::left)),
+        /// allowing it to be overridden.
+        ///
+        /// For widgets with a [`position_type`](Self::position_type) of [`PositionType`](PositionType::ParentDirected)
+        /// this acts like margin around the widget. For [`PositionType`](PositionType::SelfDirected) this
+        /// acts as the actual position from the parent.
+        pub offset: StyleProp<Edge<Units>>,
```
MrGVSV, line 171: Let me know if this should be called something else. I opted for `offset` over `position` since it can act like either margin or position depending on the `position_type`.

kayak_core/src/styles/mod.rs
```diff
@@ -138,22 +145,40 @@ define_styles! {
...
         pub min_width: StyleProp<Units>,
+        /// The positional offset from this widget's default position
+        ///
+        /// This property has lower precedence than its more specific counterparts
+        /// ([`top`](Self::top), [`right`](Self::right), [`bottom`](Self::bottom), and [`left`](Self::left)),
+        /// allowing it to be overridden.
+        ///
+        /// For widgets with a [`position_type`](Self::position_type) of [`PositionType`](PositionType::ParentDirected)
+        /// this acts like margin around the widget. For [`PositionType`](PositionType::SelfDirected) this
+        /// acts as the actual position from the parent.
```
MrGVSV, line 170: Should we also include this for the individual components as well? Specifically about the fact that their behavior is determined by their `position_type`.

StarArawn: I think this is fine for now. :)

kayak_core/src/styles/mod.rs
StarArawn: offset is fine 👍

StarArawn (approved):

PR #83, opened 2022-02-19
Split the renderer

TheRawMeatball:
Enable using the kayak_ui renderer with alternative data propagation logic


bevy_kayak_ui/src/render/font/mod.rs
```diff
@@ -17,7 +17,7 @@ mod font_mapping;
 
 use crate::BevyContext;
 
-use super::pipeline::UnifiedPipeline;
+use bevy_kayak_renderer::render::unified::pipeline::UnifiedPipeline;
 pub use extract::extract_texts;
 pub use font_mapping::*;
```
StarArawn, line 23: On line 29 you probably want to move `TextRendererPlugin` into the `bevy_kayak_renderer`? Otherwise I don't think it'll work right as a standalone renderer.

StarArawn (changes requested):
One small note that might cause issues, but as a whole this looks good! 👍

StarArawn (approved):

PR #63, opened 2022-01-15
Simple inspector widget.

StarArawn:
WIP

PR #86, opened 2022-02-21
[RFC 3] Layout Callbacks

MrGVSV:
Add a `LayoutEvent` and related props to `WidgetProps` so that widgets can be notified of layouts.

## Related
- #58
- #63 
- #84


rfcs/rfc-3-layout-callbacks.md
```diff
@@ -0,0 +1,222 @@
...
+
+As for which dispatcher to use, it might be worth creating a custom `LayoutEventDispatcher` to handle the layout events. The reason for this is purely organizational. It's probably better to separate out the concerns so that `KayakContext` doesn't get too bloated.
+
+##### Logic
+
+One complication with the logic is determining which widgets to notify. To keep it simple, we can just notify all widgets that have been re-rendered. 
+
+Another additional complication is that we can't just dispatch the event over all indices in `WidgetManager::dirty_nodes` since those only contain widgets that requested a re-render (i.e., their state updated or a binding marked them dirty). Iterating over only those would skip out on all the children, which may have had their own layouts changed after rendering.
+
+This means we actually need to use `WidgetManager::dirty_render_nodes` in order to get all re-rendered nodes. So instead of draining this collection in `WidgetManager::render()`, we'll need to keep them around and drain them after calculating the layout in `KayakContext::render()`— which we'll do in `LayoutEventDispatcher::dispatch(...)`.
```
StarArawn, line 162: Layout calculation doesn't happen until morphorm is called. The entire tree is recalculated every frame. We would need to call `on_layout` for all widgets after morphorm runs?

StarArawn (changes requested):
This seems like a big change to me, but maybe I'm just missing something. :)

rfcs/rfc-3-layout-callbacks.md
MrGVSV: I don't think we need to do it for all widgets every frame. We just need to do it for widgets that have been re-rendered (marked dirty).

So the first render will have all widgets, but subsequent ones should only concern the widgets that actually changed— assuming only updated widgets are marked as "dirty render nodes".

StarArawn: So we would need to not drain the dirty nodes and instead clear them after we call the `on_layout` after morphorm runs?

MrGVSV: Yeah I think so 🤔

StarArawn: What about children of dirty nodes? Or even parents that might have changed because a child re-rendered?

MrGVSV: The children seem to also be added to dirty render nodes (but not dirty nodes) so they should be fine. I'll run some more tests to ensure that's always the case.

But you raise a fair point about the parent widget. We probably need to add the parents of each sub-tree to the list of nodes to dispatch, which likely only requires we iterate once through the nodes to add all missing parents (it's an `IndexSet` so it will automatically prevent duplicates) before draining and dispatching.

MrGVSV: > The children seem to also be added to dirty render nodes (but not dirty nodes) so they should be fine. I'll run some more tests to ensure that's always the case.

Running the `Counter` example, it seems like every widget in the tree within the updated widget is properly added to `WidgetManager::dirty_render_nodes`. So we should be okay to iterate through that when firing the layout events.

> But you raise a fair point about the parent widget. We probably need to add the parents of each sub-tree to the list of nodes to dispatch, which likely only requires we iterate once through the nodes to add all missing parents (it's an `IndexSet` so it will automatically prevent duplicates) before draining and dispatching.

One other point to make about this, is we'll probably want to snag the _valid_ parent via `WidgetManager::get_valid_parent(...)` instead of from the normal tree.

We could try to optimize by just adding the parent of the last widget in the set (since it's an ordered `IndexSet`, the last entry is the root of the subtree). This would save a lookup on each node in the subtree. However, if we did do this, we'd need to keep in mind provider/consumer contexts. Right now, if the data is changed by a descendant, both the provider and all its children are re-rendered. But this may not always be the case— in the future we might update only the consumers rather than the whole subtree below the provider.

So is that an optimization we need/want? Or maybe we just leave that decision until after implementation when we can actually bench test and compare results. Any thoughts? 😅

rfcs/rfc-3-layout-callbacks.md
```diff
@@ -0,0 +1,222 @@
...
+    
+    for node_index in dirty_nodes {
+      // We should be able to just get layout from WidgetManager here
+      // since the layouts will be calculated by this point
+      let mut layout_event = /* ... */
+      
+      let mut widget = context.widget_manager.take(node_index);
+      let mut context_ref = KayakContextRef::new(context, Some(node_index));
+      widget.on_layout(&mut context_ref, &mut layout_event);
+      context.widget_manager.repossess(widget);
```
MrGVSV, line 189: ### Quick Note

I was just thinking that we might want to only fire this event when the layout actually changes, but that might be a bad idea. I think only firing on a change in layout might result in inconsistent behavior (at least from a user's perspective).

We could check and see what React Native does if we really want to be consistent with how they do things.

MrGVSV: Just did a quick test, and it seems React Native actually _does_ only invoke this event if there is a change in layout. Is this something we want to do? Or do we just let the user check that themselves with morphorm's `GeometryChanged` flags?

rfcs/rfc-3-layout-callbacks.md
```diff
@@ -0,0 +1,222 @@
...
+  pub z: f32,
+}
+
+impl Layout {
+  /// Returns the position as a Kayak position type (which is currently just a tuple)
+  pub fn pos(&self) -> (f32, f32) {/* ... */}
+}
+
+impl Into<Rect> for Layout {/* ... */}
+impl From<Rect> for Layout {/* ... */}
```
MrGVSV, line 78: I just saw that `Rect` is actually a type from Kayak, not morphorm. Do we want to rename that struct and its fields to match `Layout`'s? Or do we want to keep them separate?

rfcs/rfc-3-layout-callbacks.md
StarArawn: Lets wait for now.

rfcs/rfc-3-layout-callbacks.md
StarArawn: I think it makes sense to do what react native does here. Is there a performance cost to that though? 🤔

rfcs/rfc-3-layout-callbacks.md
StarArawn: Lets keep them separate for now.

rfcs/rfc-3-layout-callbacks.md
MrGVSV: > Is there a performance cost to that though? 🤔

If anything it might be more performant since we don't need to dynamically call a method on a trait object. At least, we can try it out this way then bench later to see how much of a hit it causes.

PR #87, opened 2022-02-25
Change default font

Ygg01:
- Moves `default_font` to `props`
- Add `set_default_font` 
- Change docs and examples

Stuff to discuss: 
* Is this the correct solution?
* I didn't fully understand the inheritance thingy.


kayak_core/src/styles/mod.rs
```diff
@@ -300,6 +300,8 @@ define_styles! {
         pub top: StyleProp<Units>,
         /// The width of this widget
         pub width: StyleProp<Units>,
+        /// The font style of the widget
```
MrGVSV, line 303: Might want to clarify that this expects the font name.
```suggestion
        /// The font name of the widget
        ///
        /// Only applies to [`RenderCommand::Text`]
```

kayak_core/src/styles/mod.rs
```diff
@@ -338,6 +340,7 @@ impl Style {
             row_between: StyleProp::Default,
             top: StyleProp::Default,
             width: StyleProp::Default,
+            font: StyleProp::Default,
```
MrGVSV, line 343: I think this should probably be `StyleProp::Inherit`, right? 🤔

src/widgets/text.rs
```diff
@@ -51,11 +51,11 @@ pub fn Text(props: TextProps) {
         ..
     } = props.clone();
     let font_name = font;
+    let default_style = props.clone().styles.unwrap_or_default().font.resolve();
```
MrGVSV, line 54: Hm, if we make `font` inherit, this will always panic (since we don't know what to resolve to until after style resolution)

src/widgets/text.rs
```diff
@@ -84,13 +84,15 @@ pub fn Text(props: TextProps) {
...
+
     if should_render {
         let render_command = RenderCommand::Text {
             content: content.clone(),
             size,
             parent_size,
             line_height: line_height.unwrap_or(size * 1.2),
-            font: font_name.clone().unwrap_or("Roboto".into()),
+            font: font_name.clone().unwrap_or(default_style),
         };
```
MrGVSV, line 96: I wonder if maybe we could make `RenderCommand::Text` only take the content. Then in `WidgetManager::recurse_node_tree_to_build_primitives(...)` we can actually resolve the styles and get other metadata to build the full primitive.

That might not be super optimal though since it would happen every render. 🤔

bevy_kayak_ui/src/render/font/font_mapping.rs
```diff
@@ -62,6 +62,11 @@ impl FontMapping {
         }
     }
 
+    /// Set a default `KayakFont`
+    pub fn set_default(&mut self, handle: Handle<KayakFont>) {
+        self.add("", handle);
```
StarArawn, line 67: I'd prefer to always use a non-empty string here. It might make sense to have a default font field in the font mapping instead?

src/widgets/text.rs
StarArawn: This is a good point. I wonder how CSS handles inheritance in this situation. 🤔

bevy_kayak_ui/src/render/font/font_mapping.rs
Ygg01: The only way to make this work is to change `Default` for `FontMapping`.

See 214446621c2eeb95995b6b6b5a9c9ce8eab959e1

src/widgets/text.rs
```diff
@@ -28,6 +28,20 @@ pub struct TextProps {
     pub focusable: Option<bool>,
 }
 
+impl Default for TextProps {
+    fn default() -> Self {
+        TextProps {
+            content: "".to_string(),
+            font: Some(bevy_kayak_ui::DEFAULT_FONT.into()),
```
MrGVSV, line 35: This actually won't work since `bevy_kayak_ui` is supposed to only be accessible with the `bevy_renderer` feature enabled.

But I'm looking into doing an overhaul of how fonts are handled so it might not be necessarily needed.

kayak_core/src/widget_manager.rs
```diff
@@ -406,6 +471,45 @@ impl WidgetManager {
...
     }
 
+    /// Bind a widget so that it re-renders when the binding changes
+    ///
+    /// # Arguments
+    ///
+    /// * `id`: The ID of the widget
+    /// * `binding`: the binding to watch
+    ///
+    pub(crate) fn bind<T>(&mut self, id: Index, binding: &Binding<T>)
```
MrGVSV, line 481: In a future PR, we'll probably have `KayakContext` use this method for state management rather than having its own duplicate logic for that.

MrGVSV (commented):
This is the text layout overhaul which moves the font logic into `WidgetManager`. I also added `font_size` and `line_height` to `Style`.

Right now it has a big [issue](https://github.com/StarArawn/kayak_ui/pull/87#discussion_r815689727). Initial layout isn't known until after a render, which results in layout issues until re-rendered/updated. If there are any suggestions on how to best fix this, please let me know!

## Examples

<details>
<summary>Toggle Examples</summary>

<p align="center">
text-issue-01
</p>

https://user-images.githubusercontent.com/49806985/155954940-3bf008d4-6e1e-4fc2-9d56-ab760a0e9de9.mov

</details>

MrGVSV:
This latest commit resolves the [layout issue](https://github.com/StarArawn/kayak_ui/pull/87#discussion_r815689727). To recap, the issue was that text relies on parent layout. However, if that layout has not been calculated by the time we need to render the text, then it results in weird layouts (see https://github.com/StarArawn/kayak_ui/pull/87#pullrequestreview-894788574).

To solve this, I ended up making `WidgetManager::render` recursive. If any node relies on layout to render properly, it can re-insert itself into `dirty_render_nodes`. This will tell the method that it needs to calculate layout and run again.

This extra calculation is not ideal, but it should hardly ever happen (typically only needed for use with `If`, `Fold`, etc. where widgets tend to pop in or out). Additionally, it will only allow a maximum recursion depth of `2`. Anything beyond that will need to wait until next frame.

If we don't want to do this for performance reasons, then we can just re-insert into `dirty_render_nodes` and let things work themselves out next frame. The only downside to doing that, though, is it will often create a 1-frame visual artifact/flicker as it figures out its proper layout.

kayak_core/src/widget_manager.rs
```diff
@@ -248,6 +261,58 @@ impl WidgetManager {
...
+                line_height,
+                ..
+            } => {
+                // --- Bind to Font Asset --- //
+                let asset = assets.get_asset::<KayakFont, _>(font.clone());
+                self.bind(id, &asset);
+
+                if let Some(font) = asset.get() {
+                    if let Some(parent_id) = self.get_valid_parent(id) {
+                        if let Some(parent_layout) = self.get_layout(&parent_id) {
```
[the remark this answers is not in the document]
MrGVSV: Resolved by making the `render` method recursive (up to a max of 2). While rendering, if we add to the `dirty_render_nodes`, we can force the render method to be recursively invoked. This allows all widgets that rely on some layout to be rendered properly (prevents flickering for widgets that can be properly rendered within that recursion limit— which is most widgets).

If a widget can't be rendered properly within the recursion limit, it will be handled the next frame.

kayak_core/src/assets.rs
```diff
@@ -1,7 +1,11 @@
 use std::{collections::HashMap, path::PathBuf};
+use resources::RefMut;
 
 use crate::{Binding, MutableBound};
 
+// TODO: Consider aliasing Binding<Option<T>> to be more ergonomic (or maybe use a wrapper struct)
+// pub type AssetRef<T> = Binding<Option<T>>;
```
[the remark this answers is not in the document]
StarArawn: `AssetRef` sounds good. Lets save that for another PR though.

kayak_core/src/lifetime.rs
```diff
@@ -0,0 +1,57 @@
+use crate::flo_binding::Uuid;
+use crate::{Binding, Changeable, Releasable};
+use std::collections::HashMap;
+
+/// A container for storing callbacks tied to the lifetime of a widget
+#[derive(Default)]
+pub(crate) struct WidgetLifetime {
```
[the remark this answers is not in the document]
StarArawn: WidgetLifetime sounds good to me! 👍

StarArawn (approved):

kayak_core/src/render_primitive.rs
```diff
@@ -68,6 +68,24 @@ impl From<&Style> for RenderPrimitive {
             style.border_color.resolve()
         };
 
+        let font = if matches!(style.font, StyleProp::Default) {
+            String::from(crate::DEFAULT_FONT)
+        } else {
+            style.font.resolve()
+        };
```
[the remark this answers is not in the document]
MrGVSV: Added in #98

Ygg01:
Thanks a bunch, are there examples with default fonts?

MrGVSV:
> Thanks a bunch, are there examples with default fonts? I should add some docs/examples.

Most of the examples have been converted to using `set_default` I believe. But if there's any place you feel needs better docs/examples please go ahead and submit another PR! We definitely appreciate it!

PR #89, opened 2022-03-04
[RFC 4] Text Selection Overview

MrGVSV:
## Objective

This RFC is meant to be an _overview_ of text selection. It focuses less on implementation and more on the general concepts. This way, we have a single source of truth as we implement (big changes to the design or specs after finalizing this document might require an addendum document to make note of the changes, a la #72).

### Why?

This RFC should set the groundwork for implementing things like text selection, displaying the text insertion cursor (caret), copying/cutting/pasting, etc.

Not all of those will be immediately available to implement, but will likely use features from this RFC as part of their design.

## Feedback

Discussion and feedback are very important in making sure the features proposed are generally accepted, not missing important details, and make the lives of implementors easier.

If you have any thoughts, comments, suggestions, or critiques, please leave a comment or review!

### DIFF
--- a/rfcs/rfc-4-text-selection-design.md
+++ b/rfcs/rfc-4-text-selection-design.md
@@ -262,7 +262,7 @@ Node 3 - Some("Foo") | Total - "Foo"
 
 Because Node 0 has no text content, its offset refers to children rather than characters. Because of this, ranging from an offset of 0 to an offset of 1 captures the entirety of the first child. And this captures the entire string, `"Foo"`.
 
-> In fact, we've been doing this already. A node that's fully captured by a range can be seen as sub-range from (Node, 0) to (Node, Length-1). And we can apply the same rules to the children as well, which is why we capture *all* of Node 3 in the example above. This is pretty much what happens when a range spans across a full node, such as in [Example 2.2.1](#example-2.2.1).
+> In fact, we've been doing this already. A node that's fully captured by a range can be seen as sub-range from (Node, 0) to (Node, Length-1). And we can apply the same rules to the children as well, which is why we capture *all* of Node 3 in the example above. This is pretty much what happens when a range spans across a full node, such as in [Example 2.2.1](#example-221).
 
 If instead we ended the range at (4, 1), we'd get the following traversal:
 
@@ -440,7 +440,7 @@ The *anchor* is the part of the selection that doesn't change while the *head* i
 * If the selection direction is *backwards*— the anchor comes after the head— we set the anchor as the end and the head as the start
 * If the selection direction is *directionless*— the anchor and the head are the same (a collapsed range)— it doesn't matter which is set to which
 
-Additionally, the selection should contain the *common ancestor* of the two bounds. This is the deepest node in the widget tree that contains both bounds and is vital for efficiently processing certain calculations. This will also be very helpful when we add [dynamic ranges](#2.4.-dynamic-vs-static) since it can help determine if a tree mutation possibly affected the selected contents.
+Additionally, the selection should contain the *common ancestor* of the two bounds. This is the deepest node in the widget tree that contains both bounds and is vital for efficiently processing certain calculations. This will also be very helpful when we add [dynamic ranges](#24-dynamic-vs-static) since it can help determine if a tree mutation possibly affected the selected contents.
 
 ##### 4.1.1. `Selection` Methods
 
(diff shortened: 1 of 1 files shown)


IceSentry (commented):
So, I've read about 2/3 of it and this is pretty good. My main issue though is that it seems that's going _much_ further than the stated goal of having a cursor and the ability to select text in a `TextBox`. I agree that all of these concerns will be valid at some point, but I feel like there could be a much smaller RFC that only focuses on the cursor in a single focused `TextBox`. Considering the main use of kayak right now is to make UI for games with bevy, I don't think we need something as complete as this. There's already enough complexity with this limitation, so I'm not sure we also need to be thinking about an even bigger scope.

MrGVSV:
> My main issue though is that it seems that's going _much_ further than the stated goal of having a cursor and the ability to select text in a `TextBox`. I agree that all of these concerns will be valid at some point, but I feel like there could be a much smaller RFC that only focuses on the cursor in a single focused `TextBox`.

Yeah this was one of my concerns when drafting this up. I think I really wanted this to be more of an overview of a lot of the selection-based features. Since there's a lot to consider and many parts interacting with one another, I thought it would be a good idea to go over everything all at once, then create smaller addendum RFCs that really explore one feature. But maybe that's not the best idea if it just causes confusion and is too messy. Does the "Implementation Guide" at the bottom help or would you say it definitely needs to be split up?

> Considering the main use of kayak right now is to make UI for games with bevy, I don't think we need something as complete as this. There's already enough complexity with this limitation, so I'm not sure we also need to be thinking about an even bigger scope.

Yeah this is something I keep struggling with. I keep wanting to keep things general or akin to other UI standards, but that might not be something that needs to be focused on right now. We should maybe prioritize the games/bevy side of things (will have to hear @StarArawn's thoughts on it, though).

> It's possible I simply misunderstood the purpose of this RFC and that the focus multi-node selection isn't intended to be the main focus. It's just that it was the first part so it caught my eyes.

Hey, if you misunderstood something, that's on me haha. But I think what you said makes sense and I should probably hone in the scope. For example, we could maybe stick to just allowing selection within the bounds of a single node, reduce the number of supported styles, etc. Maybe even selection itself and just focus on the caret itself.

The big question is should we keep this document as the general overview still (and create more focused sub-RFCs) or just edit it down to a small set of features? 🤔

IceSentry:
Now that I had a good night of sleep to think about it, I think it's fine if we keep this as the final goal, but I think we should either have a separate smaller RFC or just make it clear that this RFC will be done in multiple steps and that the multi node selection stuff is the end goal not the first goal. I think the issue was that the RFC pretty much started with that and it's why I felt like it was too much.

I would also rework the implementation guide to take this into consideration. 

Also, adding a table of contents might help with that, because I didn't even realize there was a multi step implementation detail at the bottom.

MrGVSV (commented):
@IceSentry I clarified that this RFC is meant to be more of an outline. 

Another point of this RFC was to cover these topics in a general sense to make it easier for people to contribute to various areas if they'd like to work on part of it. Should I include that as part of it?

rfcs/rfc-4-text-selection-design.md
```diff
@@ -788,7 +869,7 @@ Accepts any `Color` value.
 
 ## Implementation Guide
```
IceSentry, line 870: I think the only missing part here would be displaying a cursor

IceSentry (approved):
I really like the latest changes. I think the first paragraph does enough to explain that this can be done by multiple people over many PRs.

rfcs/rfc-4-text-selection-design.md
MrGVSV: Good point. The `Add selection indicators` was meant to cover that but it's not very clear and a little too broad.

rfcs/rfc-4-text-selection-design.md
```diff
@@ -0,0 +1,906 @@
...
+
+### Character
+
+Unless otherwise stated, a <em>**character**</em> is a blanket term referring to a *variable-width* character. This means that a character is not necessarily a single byte and could be made up of one or more, as is often the case when working with encodings such as [UTF-8](https://en.wikipedia.org/wiki/UTF-8). 
+
+It's important to account for this because it better prepares us for multi-language support.
+
+Additionally, when it comes to text selection, we generally want the [*grapheme cluster*](https://unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries). This is what—to a user— is typically viewed as a "character".
+
+> Rust's built-in [`chars()`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.chars) method breaks strings down into *Unicode Scalar Values*. For segmentation along grapheme clusters, an external crate like [`unicode-segmentation`](https://crates.io/crates/unicode-segmentation) will likely need to be used.
```
StarArawn, line 124: There a few different crates for managing this. It might be worth while looking into using some of those.

StarArawn (approved):

PR #90, opened 2022-03-21
Rfc 3 on layout

Ygg01:
Should implement the things needed for RFC 3.

Overall it was pretty fun and in-depth RFC, even though I'm learning the ropes of kayak_ui.

Suggestions, spelling errors, etc., are welcome


kayak_core/src/layout.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// width of the component
+    pub width: f32,
+    /// height of the component
+    pub height: f32,
+    /// x-coordinates of the component
+    pub x: f32,
+    /// y-coordinates of the component
+    pub y: f32,
+    /// z-coordinates of the component
+    pub z_index: f32,
```
MrGVSV, line 17: I think it would be better to name this just `z` since "index" implies that it's an integer.

kayak_core/src/layout.rs
```diff
@@ -0,0 +1,68 @@
...
+            x: rect.posx,
+            y: rect.posy,
+            z_index: rect.z_index,
+        }
+    }
+}
+
+///
+/// Struct used for
+///
```
MrGVSV, line 53: We should finish this doc comment

kayak_core/src/layout.rs
```diff
@@ -0,0 +1,68 @@
...
+pub struct LayoutEvent {
+    pub layout: Layout,
+    pub flags: GeometryChanged,
+    pub target: Index,
```
MrGVSV, line 57: Can we also add doc comments to these as well? I can see `flags` and `target` being confusing to some users.

kayak_core/src/layout.rs
```diff
@@ -0,0 +1,68 @@
+use crate::layout_cache::Rect;
+use crate::Index;
+use morphorm::GeometryChanged;
```
MrGVSV, line 3: Can we also re-export this using `pub use morphorm::GeometryChanged;`?

kayak_core/src/layout_dispatcher.rs
```diff
@@ -0,0 +1,65 @@
...
+        // Use IndexSet to prevent duplicates and maintain speed
+        let mut parents: IndexSet<Index> = IndexSet::default();
+
+        for node_index in dirty_nodes {
+            // If layout is not changed -> skip
+            if context
+                .widget_manager
+                .layout_cache
+                .geometry_changed(node_index)
+                == GeometryChanged::default()
```
MrGVSV, line 25: You _might_ be able to do this?

```suggestion
            if context
                .widget_manager
                .layout_cache
                .geometry_changed(node_index)
                .is_empty()
```

Alternatively, this could also be done as a `.filter(...)` after the `.drain(...)`.

Not super important either way haha, so whichever you think is clearest.

kayak_core/src/on_layout.rs
```diff
@@ -0,0 +1,52 @@
...
+use std::sync::{Arc, RwLock};
+
+/// A container for a function that handles layout
+///
+/// This differs from a standard [`Handler`](crate::Handler) in that it's sent directly
+/// from the [`KayakContext`](crate::KayakContext) and gives the [`KayakContextRef`]
+/// as a parameter.
+#[derive(Clone)]
+pub struct OnLayout(
+    Arc<RwLock<dyn FnMut(&mut KayakContextRef, &mut LayoutEvent) + Send + Sync + 'static>>,
```
MrGVSV, line 13: Ever since you brought this up, I'm debating on whether we should keep `&mut LayoutEvent` over `&LayoutEvent`.

My initial concern was consistency with `OnEvent` and future methods that require mutable access (similar to `Event::stop_propagation()`). However, I think you're right and that it might be more intuitive to users if this is immutable, so they don't try to "edit the layout".

Those "future methods" could then use some sort of interior mutability if necessary. Thoughts on that?

kayak_core/src/on_layout.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl Debug for OnLayout {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("OnLayout").finish()
+    }
+}
+
+impl PartialEq for OnLayout {
+    fn eq(&self, _: &Self) -> bool {
+        // TODO what goes here
```
MrGVSV, line 49: We can remove this TODO now
```suggestion
        // Never prevent "==" from being true because of this struct
```

kayak_core/src/widget.rs
```diff
@@ -86,6 +87,10 @@ pub trait WidgetProps: std::fmt::Debug + AsAny + Send + Sync {
     ///
     /// Returns `None` if this widget doesn't contain a custom event handler
     fn get_on_event(&self) -> Option<OnEvent>;
+    /// Gets the custom layout handler of this widget
+    ///
+    /// Returns `None` if this widget doesn't contain a custom event handler
```
MrGVSV, line 92:
```suggestion
    /// Gets the custom layout event handler of this widget
    ///
    /// Returns `None` if this widget doesn't contain a custom layout event handler
```

src/widgets/button.rs
```diff
@@ -17,6 +17,7 @@ pub struct ButtonProps {
     pub styles: Option<Style>,
     pub children: Option<Children>,
     pub on_event: Option<OnEvent>,
+    pub on_layout: Option<OnLayout>,
```
MrGVSV, line 20: Could you add this row to the table in `Button`'s doc comment?

src/widgets/text_box.rs
```diff
@@ -26,6 +26,7 @@ pub struct TextBoxProps {
     pub styles: Option<Style>,
     pub children: Option<Children>,
     pub on_event: Option<OnEvent>,
+    pub on_layout: Option<OnLayout>,
```
MrGVSV, line 29: Could you add this row to the table in `TextBox`'s doc comment?

examples/shrink_grow_layout.rs
```diff
@@ -0,0 +1,108 @@
+use bevy::{
```
MrGVSV, line 1: Can you add an inner doc comment at the top here briefly describing what this example is for?

examples/shrink_grow_layout.rs
```diff
@@ -0,0 +1,108 @@
...
+        <>
+            <Window position={(100.0, 100.0)} size={(400.0, 400.0)} title={"Grow Example".to_string()}>
+                <Text size={25.0} content={format!("Width: {:?}", layout_width).to_string()} />
+                <Button styles={Some(button_styles)} on_event={Some(grow)}>
+                    <Text size={20.0} content={"Grow".to_string()}/>
+                </Button>
+                <Button styles={Some(button_styles)} on_event={Some(shrink)}>
+                    <Text size={20.0} content={"Shrink".to_string()}/>
+                </Button>
+                <Button styles={Some(fill)} on_layout={Some(update_text)}>
```
MrGVSV, line 70: I think this would be better off as a `Background` or `Element` widget, rather than a `Button`.

examples/shrink_grow_layout.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let shrink = OnEvent::new(move |_, event| match event.event_type {
+        EventType::Click(..) => shrink_fn(width - 10.0),
+        _ => {}
+    });
+
+    let (layout_width, set_layout_width, _) = use_state!(0.0);
+
+    let update_text = OnLayout::new(move |_, layout_event| {
+        layout_event.layout.width *= 2.0;
```
MrGVSV, line 55: This should probably be removed, right? Otherwise the displayed width won't be accurate to the actual layout width.

examples/shrink_grow_layout.rs
```diff
@@ -0,0 +1,108 @@
...
+        println!("Width = {}", layout_event.layout.width);
```
MrGVSV, line 56: Maybe we say something like: `Layout changed! New Width = {}`? That might be slightly more informative.

examples/shrink_grow_layout.rs
```diff
@@ -0,0 +1,108 @@
...
+        layout_type: StyleProp::Value(LayoutType::Column),
+        background_color: StyleProp::Value(Color::new(1.0, 0.0, 0.0, 1.0)),
+        ..Default::default()
+    };
+
+    let grow_fn = set_width.clone();
+    let shrink_fn = set_width.clone();
+
+    let grow = OnEvent::new(move |_, event| match event.event_type {
+        EventType::Click(..) => grow_fn(width + 10.0),
```
MrGVSV, line 43: To help drive the point home that the displayed width is coming from the layout event itself and not the `width` state variable, maybe we add the `rand` crate as a dev-dependency and use that to determine how much to grow/shrink by (instead of just `10.0`). Thoughts?

examples/shrink_grow_layout.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    let update_text = OnLayout::new(move |_, layout_event| {
+        layout_event.layout.width *= 2.0;
+        println!("Width = {}", layout_event.layout.width);
+        set_layout_width(layout_event.layout.width);
+    });
+
+    rsx! {
+        <>
+            <Window position={(100.0, 100.0)} size={(400.0, 400.0)} title={"Grow Example".to_string()}>
```
MrGVSV, line 62:
```suggestion
            <Window position={(100.0, 100.0)} size={(400.0, 400.0)} title={"Grow/Shrink Example".to_string()}>
```
Hopefully, the font supports slashes 😅

MrGVSV (changes requested):
Overall, this is looking pretty good!

Some things I think we should do before this is merged:

* Add `on_layout` prop to most widgets (unless it doesn't apply/make sense, such as `TooltipProvider`, etc.)
  * Also add the row to the "Common Prop" table in the widget's doc comment
* If you want (not required), adding some comments in the example at key parts of the code to briefly explain layout events and how you're using them. This just helps other users learn from your example better.

kayak_core/src/layout.rs
```diff
@@ -1,8 +1,9 @@
 use crate::layout_cache::Rect;
 use crate::Index;
-use morphorm::GeometryChanged;
+pub use morphorm::GeometryChanged;
 
 /// A layout event context sent to widgets
+/// Similar and interchangeable with [Rect]
```
MrGVSV, line 6:
```suggestion
/// A layout event context sent to widgets.
///
/// Similar to and interchangeable with [Rect].
```

kayak_core/src/layout.rs
```diff
@@ -43,17 +44,20 @@ impl From<Rect> for Layout {
...
 }
 
 ///
-/// Struct used for
+/// Struct used for [crate::OnLayout] as layout event data.
 ///
 pub struct LayoutEvent {
+    /// Layout of target component
     pub layout: Layout,
+    /// Flags noting one of dimensions has changed.
```
MrGVSV, line 58: We can probably be more generic here:
```suggestion
    /// Flags denoting the layout change.
```

kayak_core/src/layout.rs
```diff
@@ -43,17 +44,20 @@ impl From<Rect> for Layout {
...
     pub flags: GeometryChanged,
+    /// the node id of the element receiving the layout event
```
MrGVSV, line 60:
```suggestion
    /// The node ID of the element receiving the layout event.
```

MrGVSV (commented):
Just some super small nitpicking

kayak_core/src/on_layout.rs
Ygg01: It's a hard issue. Each way is backward incompatible.

I'd go with the less newbie confusing option (`&LayoutEvent`).

MrGVSV: I'm on board. Let's do that!

MrGVSV (commented):
Yeah I think using `&LayoutEvent` will be easier for future users. 

Everything looks good to me, I think that last step is adding it to the other (relevant) built-in widgets.

Ygg01:
> Everything looks good to me, I think that last step is adding it to the other (relevant) built-in widgets.

Which ones?

MrGVSV:
> > Everything looks good to me, I think that last step is adding it to the other (relevant) built-in widgets.
> 
> 
> 
> Which ones?

I think pretty much any that a user might want to get layout data from (which is probably most of them haha). So `Element` and `Text` are good candidates, but `Fragment` and `TooltipConsumer` are probably not.

src/widgets/if_element.rs
```diff
@@ -27,6 +31,7 @@ pub struct IfProps {
 /// | `children`  | ✅        |
 /// | `styles`    | ✅        |
 /// | `on_event`  | ✅        |
+/// | `on_layout` | ✅        |
```
MrGVSV, line 34: I don't think we can do this for the `If` widget since it has no layout (i.e. `RenderCommand::Empty`).

src/widgets/text.rs
```diff
@@ -40,6 +40,7 @@ pub struct TextProps {
 /// | `children`  | ❌        |
 /// | `styles`    | ✅        |
 /// | `on_event`  | ✅        |
+/// | `on_layout` | ❌        |
```
MrGVSV, line 43: I feel like this one could be useful. Is there a reason it was excluded?

MrGVSV (changes requested):
Ran a few tests and it seems to be working properly, although there is a visual issue with the buttons in your example. Other than that, just a few more comments.

examples/shrink_grow_layout.rs
```diff
@@ -0,0 +1,131 @@
...
+        width: StyleProp::Value(Units::Auto),
+        height: StyleProp::Value(Units::Pixels(50.0)),
+        offset: StyleProp::Value(Edge::all(Units::Pixels(10.0))),
+        ..Default::default()
+    };
+
+    // Grow/Shrink button styles
+    let button_styles = Style {
+        width: StyleProp::Value(Units::Pixels(100.0)),
+        height: StyleProp::Value(Units::Pixels(24.0)),
```
MrGVSV, line 45: Last change I think (padding was a bit off when running):

```suggestion
        height: StyleProp::Value(Units::Pixels(30.0)),
```

MrGVSV (approved):
LGTM!

StarArawn (approved):

PR #92, opened 2022-03-25
bevy_kayak_renderer: Fix clipping not working

### LINKED ISSUE #85: Clip breaks on wide windows
## Problem

Clipping via the `Clip` widget seems to stop working when the window is resized too wide. On my machine it happens when the window is sized with a width of **`1387`** or higher:

| Width: 1386 | Width: 1387 |
| :---------: | :---------: |
| clip-1386 | clip-1387 |

This happens whether the window is sized via `WindowDescriptor` or manually resized. And from my tests, it occurs regardless of window _height_ or widget size (obviously as long as it overflows the `Clip` widget).

## Specs

**OS:** macOS Big Sur (11.6)
**rustc:** 1.58.1
**Kayak Version:** ~~4b45882cc2462df7892e0ee0d5370b50562cffaa~~ 137c421a3dcc9223676b52ecf283c0d49d4822ba (pre-release)

---

Edit: It seems I put the wrong commit for Kayak Version. That commit is after the bevy_kayak_render move which seems to have broken clipping altogether (at least on my machine).

MrGVSV:
## Objective

Fixes #85.

Clipping was broken after the extraction of the renderer due to `WindowSize` not being updated in the render app.

There was another issue present before that, though, which caused clipping to fail on wider wider windows. And upon inspection it actually fails for any position greater than its size (which often happens when a window is stretched and the elements move along with it). For example, if the clipped element had a width of `200` and an x-position of `199`, it would be fine. However, setting the x-position to `201` would result in the clip breaking.

## Solution

Update `WindowSize` and remove the code that checks if the position is greater than the size.

This seems to fix the issues, but @TheRawMeatball does this break anything for your use cases with ui4? I imagine it should be okay, but just wanted to check.

### DIFF
--- a/bevy_kayak_renderer/src/lib.rs
+++ b/bevy_kayak_renderer/src/lib.rs
@@ -27,6 +27,8 @@ pub struct WindowSize(pub f32, pub f32);
 fn update_window_size(
     mut window_resized_events: EventReader<WindowResized>,
     mut window_created_events: EventReader<WindowCreated>,
+    windows: Res<Windows>,
+    mut window_size: ResMut<WindowSize>,
 ) {
     let mut changed_window_ids = Vec::new();
     // handle resize events. latest events are handled first because we only want to resize each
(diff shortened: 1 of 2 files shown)


StarArawn (approved):

MrGVSV:
Actually, rather than removing the code, I updated it to use window size (which I think was the original intention).

MrGVSV:
Because I believe the original code was indeed a bug, I think this should be fine to merge (but let me know if I'm wrong and I'll revert).

StarArawn:
Looks good to me! 👍

PR #109, opened 2022-05-02
kayak_core: Improve `Tree` methods

MrGVSV:
## Objective

While working on creating the text selection API, I found that there were some `Tree` methods that could be very helpful. 

The first was the ability to quickly find a common ancestor between two nodes (turns out I might not need it, but discovered that after implementing it lol).

The second was having a way to distinguish whether a node comes before or after another node. This one is very important as it helps determine which way to iterate through the tree given a range of two nodes.

## Solution

Added a map to store the depth of each node in the tree.

> Originally, these depths were designed to be entirely maintained internally. However, this only worked when nodes were added from top down (which widgets are not). Therefore, the final design here allows nodes to be given invalid depths with the ability to recalculate them at a later time.

This map allows for easy calculations for certain methods, such as finding a common ancestor and determining the "order" of nodes.

Also took this opportunity to clean up `Tree` member access (and add a few missing helper methods). Before, all members of the tree were publicly accessible, which meant that a tree could quite easily fall into an invalid state. Now all members are private and previous usages have been migrated to a public (or crate-level) method.

##### Accessors

* `pub fn root(&self) -> Option<Index>`

  Returns the current root node.

* `pub fn depth(&self, index: Index) -> usize`

  Returns the depth of a given node.

* `pub fn get_children(&self, index: Index) -> Option<&[Index]>`

  Returns the children of a given node.

* `pub fn get_sibling_order(&self, index: Index) -> usize`

  Returns the index of a node amongst its siblings.

* `pub fn get_child_at(&self, index: Index, offset: usize) -> Option<Index>`

  Returns the child at a given index for a given node.

* `pub fn child_count(&self, index: Index) -> usize`

  Returns the number of children contained by a given node.

##### Utility

* `pub fn partial_cmp(&self, a: Index, b: Index) -> Option<Ordering>`

  Allows for two nodes to be compared to one another. This returns their left-to-right, depth-first traversal order (assuming they are part of the same tree).

* `pub fn get_common_ancestor(&self, a: Index, b: Index) -> Option<Index>`

  Finds the deepest inclusive common ancestor between two nodes (assuming they are part of the same tree).

##### Mutation

* `pub fn recalculate_depths(&mut self, from: Option<Index>)`

  Recalculates the depths of all nodes (from an optional root node).

* `pub fn add_children(&mut self, children: Vec<Index>, parent: Index)`

  Quickly add a set of children.

* `pub(crate) fn insert_direct(&mut self, index: Index, parent: Option<Index>)`

  Crate method for directly inserting a parent. This avoids all other logic normally associated with adding a node.

* `pub(crate) fn insert_children_direct(&mut self, children: Vec<Index>, parent: Index)`
  
  Crate method for directly inserting a set of children. This avoids all other logic normally associated with adding children.

### DIFF
--- a/kayak_core/src/context.rs
+++ b/kayak_core/src/context.rs
@@ -591,9 +591,9 @@ impl KayakContext {
 
     #[allow(dead_code)]
     fn get_all_parents(&self, current: Index, parents: &mut Vec<Index>) {
-        if let Some(parent) = self.widget_manager.tree.parents.get(&current) {
-            parents.push(*parent);
-            self.get_all_parents(*parent, parents);
+        if let Some(parent) = self.widget_manager.tree.get_parent(current) {
+            parents.push(parent);
+            self.get_all_parents(parent, parents);
         }
     }
 
--- a/kayak_core/src/event_dispatcher.rs
+++ b/kayak_core/src/event_dispatcher.rs
@@ -253,7 +253,7 @@ impl EventDispatcher {
         let mut event_stream = Vec::<Event>::new();
         let mut states: HashMap<EventType, EventState> = HashMap::new();
 
-        let root = if let Some(root) = widget_manager.node_tree.root_node {
+        let root = if let Some(root) = widget_manager.node_tree.root() {
             root
         } else {
             return event_stream;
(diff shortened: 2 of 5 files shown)


StarArawn:
@MrGVSV 
This caused some sort of infinite loop. Not sure why:
https://github.com/StarArawn/kayak_ui/issues/113

Issue #111, opened 2022-05-20
nine_patch flips the border vertically


ChristopherBiscardi:
If you replace the nine_patch image example with this image.

![metalPanel_blueCorner](https://user-images.githubusercontent.com/551247/169424165-ce684603-eb1f-4632-92a9-7504a566cfd5.png)

you can see that the border is flipped in the y direction (and that the border seems to be stretched.

<img width="1382" alt="image" src="https://user-images.githubusercontent.com/551247/169424132-72410e48-375f-41f1-9101-f990ebd09105.png">

This *also* happens with the default nine_patch image example, but it is harder to see. 

![panel](https://user-images.githubusercontent.com/551247/169424410-2f42e7cc-9d29-44d9-9f19-39b7e6589414.png)

If you focus on the bordering red/black lines you can see the top-left corner in the origin image is the bottom-left corner in the rendering.

<img width="1382" alt="image" src="https://user-images.githubusercontent.com/551247/169424394-ba635c3a-7834-4ef8-9715-bfd4ef7014af.png">

---

Additionally, the measurements seem to be off. In both cases the Edge::all needs to be doubled to fit the right size of pixels (from 30 to 60 for the new image, and from 15 to 30 for the old), and even then the borders appear stretched.

ex: new image at `Edge::all(30.0)` (which should be the right pixel count for the border).

<img width="1382" alt="image" src="https://user-images.githubusercontent.com/551247/169425389-c8ba9242-0613-4fff-b3da-cc1ca4309df3.png">

and `Edge::all(60.0)`

<img width="1382" alt="image" src="https://user-images.githubusercontent.com/551247/169425328-c921a6df-50b9-401f-9e4e-90664eb4b757.png">

The panel I was expecting looks more like this (it's a kenney asset from the space ui set)

![Sample](https://user-images.githubusercontent.com/551247/169425579-940d32f3-d48e-4873-bb45-56107f6de4a6.png)

StarArawn:
Fixed with the rewrite! 👍 Please reopen if you feel as though this is still an issue.

PR #112, opened 2022-05-21
Fix incorrect border alignment on ninepatch images

MrGVSV:
Fixes: #111

The borders were incorrectly swapped along the y-axis. This PR corrects their orientation.

This brings to light an issue with the current implementation, though, that I'm unsure how to fix. The borders of the nine patch are warped compared to the center of the image. This is likely due to the squishing of the edges' UVs so that they can fit their individual quads.

#### Source

Sample Image

#### Result

Warped nine patch output

### DIFF
--- a/bevy_kayak_ui/src/render/nine_patch/extract.rs
+++ b/bevy_kayak_ui/src/render/nine_patch/extract.rs
@@ -66,6 +66,13 @@ pub fn extract_nine_patch(
         uv_min: None,
     };
 
+    let top_uv_min_y = (image_size.y - border.top) / image_size.y;
+    let top_uv_max_y = 1.0;
+    let bottom_uv_min_y = 0.0;
+    let bottom_uv_max_y = border.bottom / image_size.y;
+    let middle_uv_min_y = border.bottom / image_size.y;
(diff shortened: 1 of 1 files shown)


StarArawn:
I'm not sure about the warping, but I suspect you are right. Ideally we should likely use an array texture here. That way we can keep UV's between 0-1 and also apply wrapping or repeating of textures.

StarArawn:
> > Ideally we should likely use an array texture here.
> 
> Not something I’m too familiar with. Should we make a separate issue for that?

Yeah! Is the warping an existing issue?

MrGVSV:
> Yeah! Is the warping an existing issue?

I haven't tested it, but I’m guessing it is. I can run a test later today to verify that it existed before this fix.

StarArawn:
> > Yeah! Is the warping an existing issue?
> 
> I haven't tested it, but I’m guessing it is. I can run a test later today to verify that it existed before this fix.

It also kinda looks like the nine patch border size is wrong. It should match the size in pixels of the most extreme corner. This is hard to do because things are split into 9 parts. So for the top with the blue "bar" it'll look wrong. Unity has different options which allow you to change how the 9 parts behave.

PR #117, opened 2022-06-09
Add support for texture atlases

mwbryant:
This pull request adds support for texture atlases and a relevant example (the sheet added is from https://opengameart.org/content/a-platformer-in-the-forest and claims no attribution necessary but I'll leave it up to you how to credit it or if you want to use a different example image).

This is a common need for game UI as many times assets in the game that appear on sprite sheets need to be represented in the UI and there appears to be no current way to do this.

The code is heavily based on the NinePatch implementation but there seemed to be no way to use it generally for generic sprite sheets before this pull request.

The end user interface may need some work shopping.  Currently the user needs to get the description of the sprite from bevy's atlas and add it to the widget themselves. I kept the dependency on bevy's texture atlas type out of the core ui system to match the style of the rest of this project.

An alternative to this might be to allow the user to specify the uv_min/max for any Image if you don't want additional code.  As of now these are hard set to None.


StarArawn:
I should have some time to review this soon. 👍

mwbryant:
The code is heavily inspired by the NinePatch implementation so I hope it's not too much effort! The only big style concern I had (which I realized after finishing the work) is that it may have been preferable to extend Image's functionality to support an upper left x/y and size with defaults of 0,0,width,height.  But that choice depends on if you think this needs a separate type and systems (which is what I did and is what bevy does to handle the same problem)

StarArawn (approved):

StarArawn:
Lets get the conflicted first and we can merge. 👍

mwbryant:
I believe I resolved them in the github editor.  It was just adding the files and use statements to mod.

PR #120, opened 2022-06-15
Use full path to Index in macros

mwbryant:
Fixes the dependency on the user needing to use Index.  Instead uses the path to index in the proc macro.  Updates every example to remove the use statement.


StarArawn (approved):

MrGVSV (approved):
Clean and simple fix. LGTM!

mwbryant:
Is this good to be merged then?

PR #131, opened 2022-07-29
Bump bevy version to 0.8.0

HeavyRain266:


MrGVSV:
I haven't sat down to test it yet, but does this actually compile and run with just these changes alone?

Also, Bevy 0.8 should be coming out in tomorrow (hopefully). Since we haven't gotten it updated already, we might as well wait for 0.8.

HeavyRain266:
> I haven't sat down to test it yet, but does this actually compile and run with just these changes alone?
> 
> Also, Bevy 0.8 should be coming out in tomorrow (hopefully). Since we haven't gotten it updated already, we might as well wait for 0.8.

No, I'd just started migration in discord thread in #offtopic on bevy server

StarArawn (approved):

PR #143, opened 2022-08-19
Pad shader structs to 16 bytes

### LINKED ISSUE #142: Shader structs aren't 16-byte aligned
When running the latest version of `kayak_ui` on the web, the app will throw an error at startup complaining that the flag `BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED` is not supported. Tracking down the [docs for wgpu](https://wgpu.rs/doc/wgpu/struct.DownlevelFlags.html#associatedconstant.BUFFER_BINDINGS_NOT_16_BYTE_ALIGNED), this flag is simply not supported by webgl

Basically, the structs used in `unified_pipeline` shader (`shader.wgsl`) aren't laid out in a multiple of 16 bytes. One possible solution would be to pad the structs with unused data to make sure they're a multiple of 16 - I was trying to get this working, but ended up hitting a knowledge gap

More information about alignment & sizes [in this Learn Wgpu article](https://sotrh.github.io/learn-wgpu/showcase/alignment/#alignment-of-uniform-and-storage-buffers) and [in the W3 spec for WGSL](https://www.w3.org/TR/WGSL/#memory-layouts)

It seems a bit odd that this wouldn't have come up before, but I'm also quite unfamiliar with wgsl and shaders in general - there may be something else that is causing this

The specific error message I get (Chrom on Ubuntu 22.04): 

Screenshot from 2022-08-19 00-24-07

Commander-lol:
Fixes #142 

This PR adds unused padding properties to the structs in `shader.wgsl` so that the struct layout is a multiple of 16 bytes, a requirement for WebGL. It also updates the `min_binding_size` to match the new minimum struct size of 16 (otherwise a different error is thrown).

I'm open to the new properties having their names changed to whatever, I figured that `_unused` is descriptive enough.

This change allows the latest version of `kayak_ui` to run successfully on the web (sans fonts, for now)

### DIFF
--- a/bevy_kayak_renderer/src/render/unified/pipeline.rs
+++ b/bevy_kayak_renderer/src/render/unified/pipeline.rs
@@ -337,6 +337,9 @@ struct QuadVertex {
 #[derive(Copy, Clone, ShaderType)]
 struct QuadType {
     pub t: i32,
+    pub _unused_1: i32,
+    pub _unused_2: i32,
+    pub _unused_3: i32,
 }
 
 pub struct QuadMeta {
@@ -376,9 +379,10 @@ pub fn prepare_quads(
 
     sprite_meta.types_buffer.clear();
     // sprite_meta.types_buffer.reserve(2, &render_device);
-    let quad_type_offset = sprite_meta.types_buffer.push(QuadType { t: 0 });
-    let text_type_offset = sprite_meta.types_buffer.push(QuadType { t: 1 });
-    let image_type_offset = sprite_meta.types_buffer.push(QuadType { t: 2 });
+    let quad_type_offset = sprite_meta.types_buffer.push(QuadType { t: 0, _unused_1: 0, _unused_2: 0, _unused_3: 0, });
+    let text_type_offset = sprite_meta.types_buffer.push(QuadType { t: 1, _unused_1: 0, _unused_2: 0, _unused_3: 0, });
+    let image_type_offset = sprite_meta.types_buffer.push(QuadType { t: 2, _unused_1: 0, _unused_2: 0, _unused_3: 0, });
+
     sprite_meta
         .types_buffer
         .write_buffer(&render_device, &render_queue);
--- a/bevy_kayak_renderer/src/render/unified/pipeline.rs
+++ b/bevy_kayak_renderer/src/render/unified/pipeline.rs
@@ -379,9 +379,24 @@ pub fn prepare_quads(
 
     sprite_meta.types_buffer.clear();
     // sprite_meta.types_buffer.reserve(2, &render_device);
-    let quad_type_offset = sprite_meta.types_buffer.push(QuadType { t: 0, _unused_1: 0, _unused_2: 0, _unused_3: 0, });
-    let text_type_offset = sprite_meta.types_buffer.push(QuadType { t: 1, _unused_1: 0, _unused_2: 0, _unused_3: 0, });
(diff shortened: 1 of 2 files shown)


bevy_kayak_renderer/src/render/unified/pipeline.rs
```diff
@@ -376,9 +379,25 @@ pub fn prepare_quads(
 
     sprite_meta.types_buffer.clear();
     // sprite_meta.types_buffer.reserve(2, &render_device);
-    let quad_type_offset = sprite_meta.types_buffer.push(QuadType { t: 0 });
-    let text_type_offset = sprite_meta.types_buffer.push(QuadType { t: 1 });
-    let image_type_offset = sprite_meta.types_buffer.push(QuadType { t: 2 });
+    let quad_type_offset = sprite_meta.types_buffer.push(QuadType {
+        t: 0,
+        _unused_1: 0,
```
StarArawn, line 384: Instead of `_unused` can we please use `_padding`? Thanks!

StarArawn (changes requested):

bevy_kayak_renderer/src/render/unified/shader.wgsl
```diff
@@ -19,6 +23,7 @@ struct VertexOutput {
     @location(3) size: vec2<f32>,
     @location(4) border_radius: f32,
     @location(5) pixel_position: vec2<f32>,
+    @location(6) _unused: f32,
```
StarArawn, line 26: I don't think this is required?

Commander-lol: Gave it a try without - it works! Is the reason for this that `VertexOutput` is only created & used inside the shader, but `QuadType` is provided from outside the shader?

bevy_kayak_renderer/src/render/unified/pipeline.rs
Commander-lol: Changed in the latest commit :+1:

bevy_kayak_renderer/src/render/unified/shader.wgsl
```diff
@@ -7,7 +7,11 @@ var<uniform> view: View;
 
 struct QuadType {
     t: i32,
+    _padding_1: i32,
```
StarArawn, line 10: Just an informative comment. Because `QuadType` is not in an array we don't actually need to specify the padding in the shader. I think it provides clarity though and matches the CPU side so we don't need to change this.

StarArawn (approved):

bevy_kayak_renderer/src/render/unified/shader.wgsl
Commander-lol: I see! Thanks for the info, it's useful to know

bevy_kayak_renderer/src/render/unified/pipeline.rs
```diff
@@ -337,6 +337,9 @@ struct QuadVertex {
 #[derive(Copy, Clone, ShaderType)]
 struct QuadType {
     pub t: i32,
+    pub _padding_1: i32,
```
StarArawn, line 340: Another option here is to use IVec3 like:
```rust
struct QuadType {
    pub t: i32,
    pub _padding: iVec3,
}
```

Commander-lol: That would be a lot cleaner - looking back through the docs, I was reading the `AlignOf` column and not the `SizeOf` column for the different types, so thought the `vec3` types were all 16 bytes

PR #147, opened 2022-10-20
Rewrite

StarArawn:


StarArawn:
Closes #53 as we directly use bevy color now! 👍

PR #150, opened 2022-10-28
Update: startup resources are now plugin settings

Laocoon7:
ImageSettings was removed from Bevy, startup settings are now plugin settings not resources. These changes allow the example code to compile with the recommended(required) versions of Bevy.

See: https://github.com/bevyengine/bevy/commit/5622d56be1262fce8a2d458f2b78abf1cab3ab50


StarArawn (approved):
I missed these thank you! 👍

PR #174, opened 2022-11-24
Implemented grid layout style props

### LINKED ISSUE #169: How to define Grid layout
I probably overlooked sth obvious but when using the grid layout, how can I define the columns and rows?

maccesch:
Fixes #169.

Unfortunately somehow after the first frame the style props row_index and col_index are unset. I couldn't find the issue and I don't think it's in my code.


StarArawn (changes requested):
This looks great, let's add an example perhaps inside of a window widget to showcase this. As for values being cleared every frame that can be due to a lot of things again an example can help me diagnose the issue. :) Also please run `cargo fmt` so CI is happy.

examples/layout.rs
```diff
@@ -0,0 +1,143 @@
...
+                        />
+
+                    </BackgroundBundle>
+                    <TextWidgetBundle
+                        text={TextProps {
+                            content: "R1 C0".into(),
+                            ..default()
+                        }}
+                        styles={KStyle{
+                            row_index: 1.into(),
```
StarArawn, line 90: This is a bit confusing and I'm hoping I can fix this in the future but text has some special requirements for styles which causes it to overwrite styles internally. The recommended way of setting styles on text is to use `user_styles` which is a field in `TextProps` I tested locally and this should solve the issue you saw. :)

StarArawn (changes requested):

examples/layout.rs
maccesch: I had the same issue with `KButton` though. Do they also have this requirement?

StarArawn: > I had the same issue with `KButton` though. Do they also have this requirement?

Not all of them but any widgets that mutate their own styles. I think what is going to happen is that widgets will be required to have a `computed_styles: KStyle` on all widgets that have styles. So that would look something like:
```rust
rsx! {
  <KButtonBundle
    styles={KStyle {
      width: Units::Pixels(240.0).into(),
      ..Default::default(),
    }}
  />
}

// In KButtonBundle
pub struct KButtonBundle {
  pub button: KButton,
  pub styles: KStyle,
  pub computed_styles: ComputedStyles,
  pub on_event: OnEvent,
  pub widget_name: WidgetName,
}

*computed_styles = KStyle::default()
  .with_style(KStyle {
      render_command: StyleProp::Value(RenderCommand::Quad),
      ..Default::default()
  })
  .with_style(styles.clone())
  .with_style(KStyle {
      background_color: Color::rgba(0.254, 0.270, 0.349, 1.0).into(),
      border_color: if state.hovering {
          hover_color.into()
      } else {
          Color::rgba(0.254, 0.270, 0.349, 1.0).into()
      },
      border: Edge::all(2.0).into(),
      border_radius: StyleProp::Value(Corner::all(10.0)),
      height: StyleProp::Value(Units::Pixels(28.0)),
      width: Units::Stretch(1.0).into(),
      cursor: StyleProp::Value(KCursorIcon(CursorIcon::Hand)),
      ..Default::default()
  }).into();
```

StarArawn: @maccesch 
Please see #179 for a fix.

maccesch: nice work! It works now! Just pushed the updated branch.

StarArawn (approved):

PR #186, opened 2022-12-02
Adding spawn during update feature

Kjolnyr:
This PR let you spawn new widgets during the Update loop. Let me know if you require some changes before merging.

EDIT: Won't work with the `rsx!` macro though. It would require a lot of work, or even a second macro to work with my changes.


Kjolnyr:
This is in fact bad. I will redo it to be more aligned with the current philosophy.

PR #187, opened 2022-12-02
Implemented spawn during update loop

Kjolnyr:
The idea behind this PR is to be able to spawn arbitrary widgets during the `CoreStage` loop. For instance, after pressing a button. I've made an example (`spawn_widget_in_update.rs`) to showcase it.

It will work with the `rsx!` syntax so this is way better than my previous PR. 

The main change is about the `spawn_widget()` function in `context.rs`. It won't check if the parent already has a child entity, so we can create multiple children. I know that there's a risk of spawning new entities during a re-render, but I think it's possible with `KayakWidgetContext`, not `KayakRootContext` since the latter is only used during the UI setup.


Kjolnyr:
Back to draft as it turns out, the `spawn_widget` function in `KayakRootContext` is now spawning new entities when it shouldn't.

I think the best way to handle this case is to have a different function when creating new widgets, and when rendering existing ones.

PR #188, opened 2022-12-02
make rsx macro return the root entity

Kjolnyr:
The idea is to be able to do something like 
```rust
let my_text_entity = rsx! {
    <TextWidgetBundle
        // ...
    />
};
``` 
it does impact all the examples as we now have to put a semicolon after each rsx! call if they are invoked with curly brackets. 

It might be better to have a macro token returning the entity only when needed?

### DIFF
--- a/kayak_ui_macros/src/widget.rs
+++ b/kayak_ui_macros/src/widget.rs
@@ -71,11 +71,12 @@ impl Widget {
                     let widget_block =
                         build_widget_stream(quote! { built_widget }, constructor, 0, false);
                     (
-                        entity_id,
+                        entity_id.clone(),
                         quote! {{
                             let parent_org = parent_id;
                             #props
                             #widget_block
+                            #entity_id
                         }},
                     )
                 } else {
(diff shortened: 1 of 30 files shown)


kayak_ui_macros/src/widget.rs
```diff
@@ -71,11 +71,12 @@ impl Widget {
                     let widget_block =
                         build_widget_stream(quote! { built_widget }, constructor, 0, false);
                     (
-                        entity_id,
+                        entity_id.clone(),
```
StarArawn, line 74: `entity_id` is copy so we don't need to clone here.

StarArawn (changes requested):

kayak_ui_macros/src/widget.rs
Kjolnyr: `entity_id` here is a `TokenStream` which does is not copy

```
error[E0382]: borrow of moved value: `entity_id`
  --> kayak_ui_macros\src\widget.rs:75:25
   |
68 |                   let (entity_id, props, constructor) =
   |                        --------- move occurs because `entity_id` has type `TokenStream2`, which does not implement the `Copy` trait       
...
74 |                           entity_id,
   |                           --------- value moved here
75 | /                         quote! {{
80 | |                         }},
   | |__________________________^ value borrowed here after move
   |
```

StarArawn (approved):

PR #209, opened 2023-01-16
Add wasm_timer dependency to allow using std::time in wasm

AdamWhitehurst:


Cargo.toml
```diff
@@ -28,6 +28,7 @@ log = "0.4"
 bitflags = "1.3.2"
 reorder = "2.1"
 resources = "1.1"
+wasm-timer = "0.2"
```
StarArawn, line 31: Wait, all you have to do is just add the dependency here? That's pretty neat. 

Can we add it only if the target is WASM? You can do that with something like:
```toml
[target.'cfg(target_arch = "wasm32")'.dependencies]
wasm-timer = "0.2"
```

AdamWhitehurst:
Sorry, I was trying to test through this branch and the PR was still a draft. wasm_timer didn't seem to work. I will reopen once I have a building sltn

PR #218, opened 2023-03-09
Update to bevy 0.10

NiseVoid:
This PR updates kayak_ui to bevy 0.10. Note that this doesn't fully utilize any of 0.10 properly yet. I also fixed some issues where the Cargo.toml didn't enable features kayak_ui depends on errors about duplicate license info. `Event` was renamed to `KEvent` to fix a name conflict with `bevy::prelude::*`

I tested all examples and tested it in my own game, sadly cargo won't let me run cargo test without crashing my system, so I'm not sure about the doc examples.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,6 @@ resolver = "2"
 authors = ["John Mitchell"]
 homepage = "https://github.com/StarArawn/kayak_ui"
 repository = "https://github.com/StarArawn/kayak_ui"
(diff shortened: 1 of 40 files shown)


NiseVoid:
Hmmm, that's interesting I didn't change anything about it afaict. Adding the `derive` feature seems to have fixed it. Might've been related to a rust update fixing some issue with incorrectly using the biggest possible feature list

PR #231, opened 2023-04-21
KButtonBundle's "styles" prop now accepts font_size

querat:
fixes the following ( discord -> crate-help -> kayak_ui )
image

- `KButtonBundle`'s `styles` prop now accepts `font_size`
  - If not specified, the previously hardcoded value is used
- Also added `height` since the hardcoded value didn't let the text expand its parent
  - Same as `font_size`, 
- Added a `KStyle` with custom `font_size` to the `simple_state` example

### DIFF
--- a/examples/simple_state.rs
+++ b/examples/simple_state.rs
@@ -1,3 +1,4 @@
+use bevy::log::LogPlugin;
 use bevy::prelude::*;
 use kayak_ui::prelude::{widgets::*, *};
 
@@ -57,8 +58,8 @@ fn current_count_render(
                         ..Default::default()
                     }}
                     styles={KStyle {
-                        font_size: StyleProp::Value(48.),
-                        height: StyleProp::Value(Units::Pixels(64.)),
+                        font_size: (48.).into(),
+                        height: Units::Pixels(64.).into(),
                         ..default()
                     }}
                     on_event={OnEvent::new(
--- a/examples/simple_state.rs
+++ b/examples/simple_state.rs
@@ -56,6 +56,11 @@ fn current_count_render(
                         text: "Click me!".into(),
                         ..Default::default()
                     }}
+                    styles={KStyle {
+                        font_size: StyleProp::Value(48.),
+                        height: StyleProp::Value(Units::Pixels(64.)),
+                        ..default()
+                    }}
                     on_event={OnEvent::new(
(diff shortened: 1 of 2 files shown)


src/widgets/button.rs
```diff
@@ -61,7 +61,19 @@ pub fn button_render(
 ) -> bool {
     if let Ok((button, styles, mut computed_styles)) = query.get_mut(entity) {
         let hover_color = Color::rgba(0.592, 0.627, 0.749, 1.0); //Color::rgba(0.549, 0.666, 0.933, 1.0);
-                                                                 // let color = Color::rgba(0.254, 0.270, 0.349, 1.0);
+
+        let font_size = if let StyleProp::Value(font_size) = styles.font_size {
```
StarArawn, line 65: You can use the style prop `resolve_or` function to make this much cleaner.

examples/simple_state.rs
```diff
@@ -56,6 +56,11 @@ fn current_count_render(
                         text: "Click me!".into(),
                         ..Default::default()
                     }}
+                    styles={KStyle {
+                        font_size: StyleProp::Value(48.),
```
StarArawn, line 60: We can use `into` for a lot of these instead of wrapping everything in `StyleProp:Value`.

StarArawn (changes requested):

src/widgets/button.rs
querat: Thank you, much less verbose indeed :)

I made a separate commit. Tell me if you'd prefer to see my changes squashed and I'll rebase my branch.

src/widgets/button.rs
```diff
@@ -128,12 +132,12 @@ pub fn button_render(
                             bottom: Units::Stretch(1.0).into(),
                             left: Units::Stretch(1.0).into(),
                             right: Units::Stretch(1.0).into(),
+                            font_size: StyleProp::Value(font_size),
                             ..Default::default()
                         }}
                         text={TextProps {
```
StarArawn, line 138: I think `TextProps` shouldn't accept a `font_size` as it should auto select the font size from the style props which casscade down the tree see:
https://github.com/StarArawn/kayak_ui/blob/main/src/styles/style.rs#L423

I don't think we should make that change in this PR though. 👍

StarArawn (approved):

PR #254, opened 2023-05-03
Do not add bevy_ui feature to bevy

NiseVoid:
Seems like the flag is not needed anymore


StarArawn (approved):
Looks good to me but I need to build and test before merge. 👍

PR #275, opened 2023-07-21
Adding more click functionality

TheTacBanana:
Adds the ability to detect Right and Middle click on widgets.

Changes are breaking due to changes of Enum names. But relatively easy to fix existing code.

Existing:
```
if let EventType::Click(..) = event.event_type {
    ...
}       
```
New:
```
if let EventType::LeftClick(..) = event.event_type {
    ...
}       
```

### DIFF
--- a/examples/avsb.rs
+++ b/examples/avsb.rs
@@ -60,7 +60,7 @@ fn current_count_render(
                         move |In(_entity): In<Entity>,
                         mut event: ResMut<KEvent>,
                             mut query: Query<&mut CurrentCountState>| {
-                            if let EventType::Click(..) = event.event_type {
+                            if let EventType::LeftClick(..) = event.event_type {
                                 event.prevent_default();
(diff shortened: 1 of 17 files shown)


ethereumdegen:
IMO this could be cleaner and more backwards compatible to add a 'button type' to CursorEvent enum and leave the EventType Enum the way it was.

ethereumdegen:
okay please see PR #276 it is built off of this one.  That PR offers the same functionality, using similar code but instead of adding new EventTypes it just adds a 'button type' prop to the CursorEvent struct.  This way the examples don't even have to change and it doesn't break anyones existing codebases but still offers all the same benefits of being able to detect which mouse button has been pressed.

TheTacBanana:
Definitely a better way of implementing it, not sure why I didn't do that myself. I'm in favor of your PR, much simpler.

PR #276, opened 2023-07-26
More click events simpler

ethereumdegen:


TheTacBanana:
Definitely agree that this is the better and simpler approach compared to mine, I extended your approach a bit to support any number of mouse buttons rather than being limited to just Left, Right & Middle. Also removes a bunch of the duplicated code.

https://github.com/TheTacBanana/kayak_ui/tree/n-mouse-button-support

Although might be taking it a bit far, not sure if anyone will ever want to detect when their forward button has been clicked on an element.

